PR #169, opened 2015-10-19
Check for valid user name, don't crash if user does not exist.

jsveiga:
Hi, sorry, we worked on the same bug simultaneously. I checked the "Sanitize all input fields" commit, and although it does not crash due to spaces in the username, it still crashes with an empty username or a non-existent username.

This patch does not crash with empty/invalid usernames, and provides feedback to the user when the username cannot be found.

It does not take care of the subreddit name, which you also sanitized on "Sanitize all input fields".

If you like the username solution, I can merge it to the current master and re-submit the pull request.


jsveiga:
Closing this one, I merged the same fixes to a more recent master, sorry...

PR #171, opened 2015-10-19
Fix crash when trying to load a non-existent user profile

jsveiga:
Hi, here's the same fix to the crash when loading an non-existent user profile; please note that this substitutes you recent sanitizing on the "go to user" inputs (but keeps the one in the subreddit input).

I'm deleting the previous pull request, as it was based on an old master (hard to keep up with you, yuo work too fast!).

best regards


ccrama:
Wow, awesome man! Will definitely be adding this in asap!

jsveiga:
Thanks! I'll try to work faster and avoid this confusion again :-)

ccrama:
Haha no worries! Means a lot when people in the community push code out :)

PR #195, opened 2015-10-20
update crashreporter

l3d00m:
add slide version & api version to auto reported crashes  
don't show authorization token


ccrama:
Awesome! That will come in handy

l3d00m:
Yeah hopefully. After we had a reported crash which looks like the user was using an old version I thought it was a good idea to include the version

PR #196, opened 2015-10-20
delete user-specific files

l3d00m:
remove files that are already in .gitignore


ccrama:
That will delete it from my files as well when I do a pull request. Instead, I'll add it to the gitignore and re-push. Will go off of this list, though! Thanks

l3d00m:
No it won't because the files are already on the .gitignore
Repushing won't delete them, that must be done manually

l3d00m:
I fucked up this branch (again). Should I make a new pr or will you fix this,  @ccrama? It is just "git rm - r folder-or-file-here".

PR #201, opened 2015-10-20
lint cleanup

l3d00m:
second try :p

Btw: Are you interested in externalizing the strings into strings.xml? I can do.


ccrama:
Actually, yeah I've needed to do that for a while now. If you are willing to do that, it would be awesome! Thank you for all the help so far :)

l3d00m:
I will at least start with it, lint reports 1708 hardcoded items so far

Alexendoo:
@l3d00m I got a few items through at a previous point in the project, but things were changing too rapidly and I kinda gave up with the new settings menu being introduced, however if it helps at all there are some items done in https://github.com/Alexendoo/Slide

l3d00m:
@Alexendoo thanks, I'll merge that in my branch

ccrama:
Holy crap that's a lot. If you don't feel up to it, no worries. Is there some sort of automated system possibly?

l3d00m:
yeah, android studio helps with this first you can search for all hardcoded items and then alt enter on hardcoded string -> and i just have to think of a name for the strings

l3d00m:
I hate merge conflicts

l3d00m:
@ccrama I'm finished with externalizing all strings from the java files, now only the xml files are missing (and that should be the smaller part)

PR #276, opened 2015-10-25
delete files that are in .gitignore

l3d00m:
but not deleted on the git remote yet. That doesn't change anything on your local files though, it just no longer uploads changes for these (user specific) files


ccrama:
Awesome! Thank you!

ccrama:
Didn't mean to close :x

PR #301, opened 2015-10-27
externalize 2 more strings and other fixes

l3d00m:
clean code (move time relevant code into timeutils)
readd gitreporter token (removed with last pr)
add version code to gitreporter


SolarAquarion:
we really should be using one of those free CI's to make sure it builds

l3d00m:
That would be a good thing to do.

PR #358, opened 2015-11-04
add newest translations, add crowdin badge

l3d00m:


ccrama:
Great! Thank you very much :)

l3d00m:
Because it's mostly the same than the original strings.xml. If strings are not translated in the country specific values folder, the default strings.xml will be used.

PR #386, opened 2015-11-06
Don't allow comments with 0 childs to collapse

Spittie:
I find that quite annoying.

I've tried this for a bit and it doesn't seems to introduce any problems.
I don't know if this is the way you want to go, but I rather have a comment do nothing on tap than just put a "+0" label to it.

As a side note, I can't really understand what the whole isClicking branch or clickpos do - I've left them, but I can't find them used anywhere. It's for future usage?


ccrama:
Awesome, thank you! And yes, I had added that for a feature I was going to add but then removed it last minute. Will hook into them later!

PR #435, opened 2015-11-08
This fixes #423

### LINKED ISSUE #423: App crash: Reorder your pins view
Slide crashes if you have no pinned subs to start with. Adding pinned subs first from the sidebar info fixes the issue.

Goto:
Profile > Reorder your pins 
Tap the add button/fab and the app will crash.

Also it is not obvious that you need to pin apps from the sidebar first in order to reorder them. It would be better to have the ability to reorder all subs even if they are not pinned.

Ph0ndragX:
Now proper dialog is shown when there are no pins.


ccrama:
Good find! Thanks :D

PR #437, opened 2015-11-08
Score followed by "points" in layout

Verzwei:
Have the score being followed by points instead of having the number alone, it looks better. Example:

RokutenVerzwei 5 points 10 minutes ago

Sorry if anything there's any error, GitHub doesn't seem much Mobile friendly.


ccrama:
That actually is just a test layout, that doesn't affect the actual code :(. The real code can be found in the PopulateSubmissionViewHolder class

ccrama:
If you would like to edit that, feel free to :)

Verzwei:
Oh, right lol Sorry, it's my first venture on Mobile Apps! I'm so sorry!

PR #447, opened 2015-11-08
This fixes #422

### LINKED ISSUE #422: Crash when tapping on the username of a deleted comment
Normally, tapping on the username in a comment brings you to the profile of that user. However, if the comment has been deleted, tapping the ``[deleted]`` username will cause a crash.

Ph0ndragX:
When trying to get information about user with invalid username, the dialog is shown and then it backs to previous screen.


Ph0ndragX:
It matches any alphanumeric character, see:
http://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html
Although, I have checked what characters are actually permitted in reddit username and created fix.

l3d00m:
i fixed it already, thanks!

PR #457, opened 2015-11-08
It fixes #449

### LINKED ISSUE #449: Dots in app tutorial not working as intended
At the moment, in the tutorial, if you tap on one of the dots, represeting the slides, it will show as it's selected, but slide will stay the same. Since when you swipe to another slide the dot changes, I guess that selecting a dot should change the slide. If the issue is unclear I can add a video later.

Ph0ndragX:
Now clicking on a radio button displays a corresponding screen.


ccrama:
Awesome! Thank you!

PR #488, opened 2015-11-10
Added an option in the theme settings to enable/disable FAB

TomerRosenfeld:
FAB will be visible as long as Reddit.fab is true


SolarAquarion:
Thank You Chaver.

Issue #486, opened 2015-11-10
Profile search enter key


2fst4u:
When searching profiles, the keyboard's "enter" key is set to carriage return, rather than "go". So if I type a profile name and hit enter, a new line is entered in the field. "Go" would make more sense.

l3d00m:
Fixed a while ago

PR #489, opened 2015-11-10
Optimized imports + IDEA auto inden and solved issue #486

TomerRosenfeld:
Simply used IDEA(intellij / Android Studio) auto indent/reformat to make the code look a bit cleaner + remove unused imports.
For example
int i=5;
Becomes 
int i = 5;

class ClassName{
Becomes
class ClassName {

etc..


ccrama:
Wow that's a ton of changes. Do you mind splitting the last edit (8b310df) into a seperate PR and let me take a look through all the changes first (to make sure nothing broke)?

Thanks!

TomerRosenfeld:
Sure, how do I do that?
I'm not very experienced with git

PR #492, opened 2015-11-10
hide button, scroll to top, various fixes

### LINKED ISSUE #434: Request for hide button in action bar on comment view too
As above. Simple button to action bar instead of having to open the menu to hide a post.

### LINKED ISSUE #474: Add a way to quickly return to the top of a comment thread
Sometimes it is necessary to return to the top of a long comment thread to check out the original post again or to refresh the contents of the thread for new answers. Either an icon in the top bar or in between the previous-next comment arrows (or something else) would be great.

l3d00m:
add setting to move hide button to action bar( fixes #434)  
remove empty file  
long clicking the comment arrow (up) navigates to top, fixes #474)  
add the secret constants back!

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Activities/SettingsGeneral.java
+++ b/app/src/main/java/me/ccrama/redditslide/Activities/SettingsGeneral.java
@@ -68,6 +68,19 @@ public class SettingsGeneral extends BaseActivityNoAnim {
             });
         }
         {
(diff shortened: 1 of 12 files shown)


Asadron:
What is this SecretConstants thing? It breaks compilation for me when II have it in.

l3d00m:
That are API-keys that should not be public, means only ccrama has them. Just comment it out or set it null on your fork. It breaks only in app purchase for you.

PR #504, opened 2015-11-11
Merged most of the overviews classes into the base overview class

rosenpin:
Merged most of SubredditOverview and SubredditOverviewSingle into the OverviewBase class for easier code maintenance with less copy-pastes.


ccrama:
That was badly needed. Thank you very much!

PR #505, opened 2015-11-11
I forgot to add the secret constants...

rosenpin:


ccrama:
Hehe, no worries

Tunous:
Perhaps add public SecretConstants file with fake values and special one disabling iab if it's a trouble for contributors?

l3d00m:
that would be a good idea. it's somehow possible to  realize that though gradle flavours too

PR #528, opened 2015-11-13
Fixed bug where the toolbar would overlap the content in the overview

rosenpin:
There was an issue with one of the latest commits that caused this bug, here are screenshots of before and after the fix.

Before
After


ccrama:
Oh sweet, was about to work on this! Thank you! Also, are you on our IRC channel?

rosenpin:
No problem.
And no, I am not.

ccrama:
You should! We hang  out there a lot. The server is snoonet and the channel is #slide if you'd like to join :)

rosenpin:
I will, thanks!

PR #532, opened 2015-11-14
Instant read

rosenpin:
Posts will now change color to "read" color instantly
Here is an [example](https://youtu.be/5gLAKkSAiss)


ccrama:
Love it!

PR #539, opened 2015-11-15
Added an option in settings to use fab to dismiss viewed posts like in Relay

rosenpin:
Settings
Activity


ccrama:
Wow, this is nice! :D

PR #543, opened 2015-11-15
Fixed the issue with last PR where only some posts would be removed

rosenpin:


ccrama:
Didn't catch you on IRC. Thank you and good luck on the test!

rosenpin:
Thanks!

PR #554, opened 2015-11-17
Share link from fullscreen mode

ihatetothink:
The strings are not translated.
I took the icons from existing ones, but i think they match the meaning.

screenshot_2015-11-17-01-07-27


ccrama:
Great addition! Thank you very much!

Alexendoo:
Does this mean that the external storage permission is no longer required?

PR #555, opened 2015-11-17
Fix theme colors and icons

recalculated:
* Corrects the light/dark theme colors to the following values (specified in the MD specs):

| Theme | Light | Dark |
| --------- | --------- | --------- |
| Text | 87% black | 100% white |
| Icons | 54% black | 70% white |
| Background | #FAFAFA | #303030 |
| Cards | #FFFFFF | #424242 |

* Makes the icon drawable resources the same color

* Makes AMOLED black theme background actually black

* Changes icons for layout/alt layout to something more representative of cards

* Changes the backup icon to a cloud upload icon

* Provides density-specific versions of the hide icon

* Removes icons in drawable that have other versions in drawable-*density*. They are never used and only serve to increase APK size


ccrama:
Wow, lots of hard work went into this one. Looks VERY nice! Thank you!

I changed the activity background to be a teensy bit darker, as it was hard to distinguish cards on the background, but everything else is spot on. 

Thank you!

PR #556, opened 2015-11-17
Fix Gradle plugin version (1.3.1)

recalculated:
See http://developer.android.com/tools/revisions/gradle-plugin.html
Gradle plugin v1.5+ does not exist (1.3.1 is latest), and using an incorrect version number can cause Gradle sync errors on rare occasions.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -4,7 +4,7 @@ buildscript {
         jcenter()
     }
     dependencies {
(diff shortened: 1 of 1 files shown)


ccrama:
Thank you again for all of these PRs! :)

recalculated:
No problem. Happy to help

Issue #561, opened 2015-11-17
Gilded posts do not have any indication of being guilded (missing gilded icon)


karlding:
Gilded comments (Picture 3) will show the gilded star icon, but if the Original Post gets gilded, no indication is shown (Pictures 1 and 2).

This is the [post](https://www.reddit.com/r/gaming/comments/3t0bz1/mrw_skyrim_says_you_cant_fast_travel_when_enemies/) in the screenshots

![Skyrim](http://i.imgur.com/X5aWauN.png)

![Skyrim 2](http://i.imgur.com/NOmTPci.png)

![Skyrim Comment](http://i.imgur.com/LnD6uuM.png)

l3d00m:
fixed in #562

PR #562, opened 2015-11-17
Gild on main view 561

ihatetothink:
I have implemented what is described here: https://github.com/ccrama/Slide/issues/561 , here is what it looks like: 
screenshot_2015-11-17-12-48-46

Right now I'm unable to test on a more modern device (I'm testing at 4.4)

Maybe it's quite unaligned.. I don't know where to align it..


ccrama:
Very cool! Thank you very much!

Issue #565, opened 2015-11-17
Wrong arrow colouring when upvoting and downvoting a comment


karlding:
The arrow colours are not reset when one action is performed, and then the other.

For example, if you accidentally downvoted a comment, and then change your vote to an upvote instead, you'd expect for the downvote arrow to update to the default colour, and the tapped arrow to update to the active colour. Instead, both become coloured.

PR #567, opened 2015-11-17
Fixes the upvote/downvote colouring issue (ticket #565)

karlding:
Simple 2 liner :)


ccrama:
Wow, that was easy! Thank you

Issue #487, opened 2015-11-10
Markdown: Code Formatting


karlding:
The Markdown code formatting doesn't render at all (inline and block). Code blocks have formatting that may be tricky to deal with on smaller screens, but inline code shouldn't have that issue.

Ideally it'd switch to some monospaced font or something, to differentiate it from the normal text. Currently it looks exactly the same as regular text.

## Example Inline Markdown Code
```markdown
The line ``std::cout << "Hi" << std::endl;`` will output "Hi" to standard output.
```
Should render as "The line ``std::cout << "Hi" << std::endl;`` will output "Hi" to standard output." Instead, the ``std::cout << "Hi" << std::endl;`` is rendered just like the other text.

PR #575, opened 2015-11-18
Add monospace formatting to code fragments (issue #487)

Deadleg:
The general strategy is to take the raw HTML such as

blah <code>x = 1</code> cats

and map it to

blah <code>[[&lt;[x = 1]&gt;]]</code> cats

Then apply Html.fromHtml, which produces

blah [[<[x = 1]>]] cats

Using these tokens, the text can be parsed and a monospace font can
be applied.

The transform of the CharSequence to a string in the setCodeFont
method is to simplify the process of search for the tokens, though
it could probably be implemented using the mutable nature of
SpannableStringBuilder. Performance wise there is probably not a huge
hit since we are parsing relatively small amounts of text.

See issue #487.


ccrama:
Very cool! Thank you for this! I sidnt knoe how to get code blocks to work before, so thank you for the information

PR #587, opened 2015-11-18
Fixed multi fab style and colors inconsistency

rosenpin:
This PR also includes the other PR commit and I'm not sure how to remove it..


l3d00m:
For the next time: Use two different branches.  
It should be no problem this time, because both commits looks good!

rosenpin:
Thank you.
I used two different branches and used 'git checkout -b barnches names' to toggle between them but the local code was pushed to all of them...

l3d00m:
Yeah, that thing makes a new branch from the one you are currently using. First do 'git fetch --all' to get the newest changes from this repro then do 'git checkout - b branchname ccrama/master'. (or upstream/master). This makes a new branch from this repros master branch instead of the branch you are currently working on!

rosenpin:
Got it, thanks, will do next time.

PR #594, opened 2015-11-18
If there is no internet connection the app will close after showing the dialog

rosenpin:
Until now if there was no internet connectoin the app would show a dissmisable dialog but would keep the loading screen visible even though it won't load.


ccrama:
Sweet, thanks!

PR #595, opened 2015-11-18
fixes #573

### LINKED ISSUE #573: Settings cosmetic issues
* Text wraps unnecessarily
* There appears to be two touch targets going on
* The done button spans the entire dialog, should be on the right hand side

l3d00m:
update strings
fix fab settings crash
add back arrow to all settings activities

this fixes #573 (the part that @ccrama haven't fixed yet):
disable double material ripple
align done button to the right


ccrama:
Thanks!

PR #599, opened 2015-11-18
Fixed force close when opening an item from inbox

rosenpin:
I am not sure about the reason for this issue but it happened to me a 100% of the times I've tried to open a comment from my inbox,


ccrama:
Good find. Thanks!

PR #616, opened 2015-11-19
Added circular reveal enter transition animation

rosenpin:
[That is how it looks](https://www.youtube.com/watch?v=gySHV5mSZVQ)


ccrama:
Looks great! :D

PR #618, opened 2015-11-19
Tint navigation bar

nadeemsultan:
Tint the navigation bar to match the status bar. This is one of the features I missed from other clients. It also helps prevent some navbar burn-in on amoled displays if you spend most of your time on Reddit.

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Activities/CommentsScreen.java
+++ b/app/src/main/java/me/ccrama/redditslide/Activities/CommentsScreen.java
@@ -54,6 +54,7 @@ public class CommentsScreen extends BaseActivity {
             if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
                 Window window = getWindow();
                 window.setStatusBarColor(Pallete.getDarkerColor(posts.get(firstPage).getSubredditName()));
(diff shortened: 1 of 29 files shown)


recalculated:
According to the [developer docs](http://developer.android.com/training/material/theme.html#StatusBar), you shouldn't do this.

>When you customize the navigation and status bars, either make them both transparent or modify only the status bar. The navigation bar should remain black in all other cases.

I think it would be best to implement this as an optional setting (turned off by default).

Alexendoo:
And as an aside the navigation bar is black meaning it wouldn't need to be lit on OLED displays, meaning no burn in anyway

recalculated:
I think he's talking about 'reverse' burn-in, i.e. the rest of the screen gets burned in while the nav bar doesn't, which causes that portion of the screen to become unusually vibrant in color. It's noticeable when the screen is rotated

nadeemsultan:
Ah fair enough. And maybe burn in was the wrong word. However, the rest of the display will degrade faster than the navbar since it's not lit, so there will still be a noticeable difference between the two.

ccrama:
Will edit this to make it a setting!

l3d00m:
@nadeemsultan it should now be possible to do this in the baseactivity, which would make this easier.

Issue #610, opened 2015-11-19
On a fresh install of Slide, the default selection for comment sort (on the action bar inside posts) is [Top], when it actually isn't


okungnyo:

l3d00m:
fixed with #623

PR #623, opened 2015-11-20
Fix the comment sorting to reflect reddit defaults

karlding:
This commit changes the default sorting in the preferences to "Best" (issue #610), which is the ``Confidence`` option for ``default_comment_sort`` defined in the reddit Oauth API . In addition, an extra ``>`` in the ``Q&A (AMA)`` string resource in ``strings.xml`` was removed (not sure how that got in there in the first place).

Now the default sorting is set to ``confidence``, and logging in should keep whatever sorting option you selected.


SolarAquarion:
thanks

PR #624, opened 2015-11-20
Fix guilded post icon alignment

karlding:
These changes realign the icon added in #562, so that it lines up with the positioning of the other layout elements. Currently it looks a little out of place, being all the way off to the left. I didn't see an issue, but I figured that since this is a minor cosmetic change, I'd just create the pull request right away.

(I think these are all the layout files that needed to be modified)


SolarAquarion:
thanks

PR #634, opened 2015-11-21
Slim down Play Services dependency

recalculated:
See https://developers.google.com/android/guides/setup

Rather than including the massive play-services library, it is possible to pick out only the parts you need as dependencies. This app only uses the Google Drive functionality, so play-services can be replaced with play-services-drive. This results in a large decrease in APK size.

Before:
```
-rw-r--r-- 1 owner owner  11M Nov 21 10:55 app-debug.apk
-rw-r--r-- 1 owner owner  11M Nov 21 10:55 app-debug-unaligned.apk
```

After:
```
-rw-r--r-- 1 owner owner 9.3M Nov 21 11:08 app-debug.apk
-rw-r--r-- 1 owner owner 9.3M Nov 21 11:08 app-debug-unaligned.apk
```

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -73,7 +73,7 @@ dependencies {
     compile 'com.github.rey5137:material:1.2.1'
     compile 'jp.wasabeef:recyclerview-animators:2.0.0'
     compile 'com.sothree.slidinguppanel:library:3.2.0'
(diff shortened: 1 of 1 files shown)


ccrama:
Wow really great find. Thank you very much for these great PR's!

Issue #636, opened 2015-11-22
Severe formatting issues regarding inline code


recalculated:
Reference post: https://www.reddit.com/r/archlinux/comments/3tmmcx/makepkg_illegal_instructions_on_everything/cx7khmn

This is the actual post text:

>2 things:
>
>Can you post your `/etc/makepkg.conf`/`~/.makepkg.conf`/`~/.config/pacman/makepkg.conf`?
>
>When you run makepkg, scroll up from where the error occurs, and look for a line containing `cc` or `gcc`. Your PKGBUILD likely calls `make`, and `make` prints the commands as it runs them. Let's see the exact gcc command that it is running.

And this is what Slide displays:

![screenshot_2015-11-21-20-33-28](https://cloud.githubusercontent.com/assets/13721872/11321918/33b8856e-908f-11e5-8b85-6426d5bba735.png)

l3d00m:
fixed with #638

PR #638, opened 2015-11-22
Fix formatting issue with adjacent code blocks (issue #636)

Deadleg:
When the "[[<[" sequence is removed from the message, the position
pointer is not moved back 4 spaces to compensate for this. This
leads to 4 characters being missed on the next iteration. If the
"[[<[" sequence lies in this space, it will be missed.

Fixes issue #636


ccrama:
Thank you very much!

PR #641, opened 2015-11-22
Fix the light_green accent colour

### LINKED ISSUE #640: Selecting the light_green accent colour causes a crash
If the ``light_green`` accent colour is selected, Slide will crash, and will subsequently crash when attempting to restart the app.

karlding:
This fixes #640. Turns out the colour setting was set to an invalid resource. :astonished: (looking through the commit history, it looks like it was always set that way)


ccrama:
Wow, good find! Have never noticed that!

Issue #654, opened 2015-11-23
crash navigate in comments: next-button when comments = 0


c727:
^

l3d00m:
it's fixed with #656

PR #656, opened 2015-11-23
Fix RecyclerView IndexOutOfBounds crash

karlding:
If I'm reading the code right, this should be a simple fix for #654 to prevent an ``IndexOutOfBoundsException`` from occurring.


ccrama:
Merged it before looking, why not just do a null or empty check on the adapter? I don't think that fix will work.

l3d00m:
it works though

ccrama:
Would that not offset which comment it scrolls to though?

l3d00m:
nope, it all works like it should

ccrama:
Huh, ok! Thanks!

PR #657, opened 2015-11-23
Fine tune intent handling regex

recalculated:
* Support all reddit subdomains
* Make regex case insensitive
* Fix wiki links
* Add a short description of each regex

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/OpenRedditLink.java
+++ b/app/src/main/java/me/ccrama/redditslide/OpenRedditLink.java
@@ -18,12 +18,7 @@ public class OpenRedditLink {
     public OpenRedditLink(Context context, String url) {
         boolean np = url.matches("(https?://)?np.reddit.com.*");
 
-        url = url.replace("np.", "");
-        url = url.replace("www.", "");
-        url = url.replace("http://", "");
-        url = url.replace("https://", "");
-        url = url.replace("m.", "");
-        url = url.replace("i.", "");
+        url = url.replaceFirst("(https?://)?[a-z0-9-_]+\\.reddit\\.com", "reddit.com");
 
         if (url.startsWith("/")) url = "reddit.com" + url;
         if (url.endsWith("/")) url = url.substring(0, url.length() - 1);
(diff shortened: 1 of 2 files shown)


app/src/main/java/me/ccrama/redditslide/OpenRedditLink.java
```diff
@@ -15,58 +16,49 @@
     private final static String TAG = "OpenRedditLink";
 
     public OpenRedditLink(Context context, String url) {
-        boolean np = url.contains("np.");
+        boolean np = url.matches("(https?://)?np.reddit.com.*");
+
         url = url.replace("np.", "");
```
Alexendoo: Worth removing all the language subdomains too rather than just `np.`? regex would be `([_A-Za-z0-9]{2}-)?[_A-Za-z0-9]{2}\\.`. eg en-us.reddit.com, de.reddit.com, though technically anything even __-__.reddit.com is a working URL

recalculated:
Good catch, thanks.

Alexendoo:
I think this way still leaves a couple gaps, namely the fact that reddits subdomains are a bit crazy. It now captures all reddit subdomains if I'm reading correctly? There are some unique ones such as blog.reddit.com and store.reddit.com that lead to other pages, as well as anything in the form subreddit.reddit.com being a redirect to reddit.com/r/subreddit. Not sure if these are handled

recalculated:
subreddit.reddit.com links are handled properly (which is why I expanded the regex to allow any subdomain). I'm not sure how to handle the unique subdomains - is it possible to exclude hosts in an intent filter?

ccrama:
Very good, thank you! @Alexendoo  @recalculated  I think that the only subdomains that link to actual posts are np.reddit.com, reddi.it, www.reddit.com, i.reddit.com, and m.reddit.com. We could put a hook in there and if the data before the first period is not equal to www, "", np, i, or m, return as a link and we can just open that in the browser. 

Reddit has a few actual subdomains they use, which can be seen here http://www.wolframalpha.com/input/?i=reddit.com (click "subdomains" and Wolfram will show a list of used ones).

If the subdomain is not equal to any of those above, Reddit actually links you to that subreddit (android.reddit.com links to reddit.com/android), so we could put up a filter that checks for any of the above, and defaults to opening that subreddit if it exists.

PR #662, opened 2015-11-25
Handle subreddit.reddit.com links

recalculated:
Changes subreddit.reddit.com links to reddit.com/r/subreddit.

Side note: as @Alexendoo pointed out, there are certain reddit subdomains that this app isn't built to handle (namely blog, store, pay, and ssl). I personally don't know how to exclude these subdomains from the intent filter, so I've added a fixme to the Android manifest in case someone else does.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -122,6 +122,7 @@
                 <category android:name="android.intent.category.DEFAULT" />
                 <category android:name="android.intent.category.BROWSABLE" />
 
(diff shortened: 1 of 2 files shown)


ccrama:
Thank you very much! And yeah, I'm compiling a list of real subdomains (pay, etc), to add to a filter and just open those in browser like normal. Thank you!

ccrama:
There was a slight issue with the regex, you had //w* (alphanumeric and _) in the check for the end value (see if there is context to a permalink), but the "?" in the context is not alphanumeric. I made it .* instead, and it now works!

Fixed with 294b3c1

PR #665, opened 2015-11-25
Moved comments settings from general to its place

rosenpin:


l3d00m:
+1, that looks nice

l3d00m:
Everything works as it should, so i will merge it. Thanks for your pr!

PR #667, opened 2015-11-25
Added an option to change animation length

rosenpin:
Added a seekbar in general settings to control the open animation length


ccrama:
Really great, thanks!

PR #670, opened 2015-11-26
Use digits attribute instead of InputFilter for subreddit search (fixes #143)

### LINKED ISSUE #143: Sanitisation issues with the 'go to subreddit' input field
I cant type in underscores!

recalculated:
Applies the `android:digits` attribute to the subreddit search `EditText` rather than using a custom `InputFilter`.


Alexendoo:
Interesting, it works somewhat however it seems a little glitchy. On my end attempting to input an upper case character messes up the rest of the string, typing a few letters has a lot of duplicates for whatever reason. Hitting the spacebar by accident seems to have an effect also, it adds an empty character or something on the end if you type some characters afterwards. (&lt;space&gt;abc, you have to backspace twice to delete c)

If it helps at all I notice the log is filled with a bunch of `E/SpannableStringBuilder: SPAN_EXCLUSIVE_EXCLUSIVE spans cannot have a zero length`

recalculated:
No uppercase input was just a silly mistake on my part, I didn't realize the characters were case sensitive. Keyboard autosuggestions are the cause of duplication and other weird behavior; when I disabled suggestions in the EditText everything worked smoothly, but I'm not sure if it's the right course of action.

ccrama:
Really great work, can't wait to test it out when I'm back on my computer! Thank you!

Alexendoo:
Seems to work pretty well now it doesn't mind the case, the small issues it has left I've noticed are also present in reddit is fun and nobody complains about that one

PR #673, opened 2015-11-26
Moved secret constant to properties file

### LINKED ISSUE #558: Remove "SecretConstants" file, or dynamically enter developer mode if it isn't found.
Please remove this file and put the public key in directly, or add some other mechanism for collaborators to easily compile your project.

The google developer guidelines say "To keep your public key safe from malicious users and hackers, do not embed it in any code as a literal string. Instead, construct the string at runtime from pieces or use bit manipulation (for example, XOR with some other string) to hide the actual key. The key itself is not secret information, but you do not want to make it easy for a hacker or malicious user to replace the public key with another key." [http://developer.android.com/google/play/billing/billing_best_practices.html]

However, this is to stop people from putting their own public keys and unlocking features without paying for them.  In your case, it is trivial to simply extract the dex file and get the constant from the file.  It's even easier to just compile the app with all IAPs purchased, since this project can be compiled by anyone.

I can understand if you want to keep it private, but if you do go that route, you should add a case in the code where no key was found so that the project can still compile.

Deadleg:
To get this to work, create the file

/app/src/main/assets/secretconstants.properties

with contents

base64EncodedPublicKey=asdkajsnkasi2h2379r1

(but with the actual key of course). If the file is not present,
the constant is set to an empty string.

Fixes #558.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,4 +26,4 @@ Thumbs.db
 cd
 
 
-/app/src/main/java/me/ccrama/redditslide/SecretConstants.java
+/app/src/main/assets/secretconstants.properties
--- a/app/src/main/java/me/ccrama/redditslide/Reddit.java
+++ b/app/src/main/java/me/ccrama/redditslide/Reddit.java
@@ -452,7 +452,7 @@ public class Reddit extends MultiDexApplication implements Application.ActivityL
         @Override
         protected Void doInBackground(Void... params) {
             try {
-                mHelper = new IabHelper(Reddit.this, SecretConstants.base64EncodedPublicKey);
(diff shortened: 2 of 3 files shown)


ccrama:
Will look into this when on back on my computer later this week! Thank you!

SolarAquarion:
@ccrama have you forgotten

okungnyo:
Or ignored...

l3d00m:
Spreading a bit of negativity here? It's hard to keep track of all issues and pull requests, especially if you have a busy real life.

okungnyo:
Just a suggestion of possibility, but yeah, I see how things could get lost in the shuffle of not only developing a huge project but also managing an issue tracker.

I apologize...

ccrama:
Sorry this took so long @Deadleg . Thank you for the commit!

PR #679, opened 2015-11-27
Moved hide post button checkbox from general to layout settings

rosenpin:


app/src/main/res/layout/activity_settings_theme_card.xml
```diff
@@ -182,6 +183,42 @@
...
+                    android:layout_width="56dp"
+                    android:layout_height="match_parent"
+                    android:paddingBottom="16dp"
+                    android:paddingRight="16dp"
+                    android:paddingTop="16dp"
+                    android:src="@drawable/hide"
+
+                    android:tint="?attr/tint" />
+
+                <CheckBox
```
l3d00m: Why have you replaced it by a checkbox?

rosenpin: Because all the toggled settings in this settings screen are checkboxes

PR #680, opened 2015-11-27
Added option in general settings for fullscreen view submissions links

rosenpin:


app/src/main/java/me/ccrama/redditslide/Activities/FullscreenImage.java
```diff
@@ -36,7 +36,7 @@
 /**
  * Created by ccrama on 3/5/2015.
  */
-public class FullscreenImage extends BaseActivity {
+public class FullscreenImage extends FullScreenActivity {
```
l3d00m: Why extending it from a new class when you already checked for it in line 51-54?

l3d00m:
this includes #679

ccrama:
Thank you so much for you guys' work! Really sleek! Thanks

app/src/main/java/me/ccrama/redditslide/Activities/FullscreenImage.java
rosenpin: So it won't have to be written in every full screen activity but in the full screen activities only

PR #684, opened 2015-11-28
Issue #34, option to hide status bar in general settings

### LINKED ISSUE #34: Translucent status bar

rosenpin:
Added option in general settings to hide status bar.

fixes #34


okungnyo:
Awesome work!

app/src/main/java/me/ccrama/redditslide/Activities/SuperBaseActivity.java
```diff
@@ -0,0 +1,37 @@
...
+        translucent();
+    }
+
+    @Override
+    protected void onResume() {
+        super.onResume();
+        translucent();
+    }
+
+    public void translucent() {
```
l3d00m: You should probably do it only if Reddit.translucentStatusBar is true.

l3d00m:
We should also extend the BaseActivity in the future so that it's a real base activity.

app/src/main/java/me/ccrama/redditslide/Activities/SuperBaseActivity.java
rosenpin: You're absolutely right.

ccrama:
What's the difference between BaseActivity and SuperBaseActivity?

rosenpin:
Base activity isn't really a base activity, or adds swipe back and some more things,  superbaseactivity extends app compact activity and doesn't add anything but app wide features.

recalculated:
I think it would be better to revise the BaseActivity so that all activities can extend it, and then implement the feature in BaseActivity. It doesn't really make sense to have a BaseActivity that isn't really a base, and then create another SuperBaseActivity that does something that should just go in the actual base activity.

l3d00m:
@rosenpin BaseActivity is now a real baseactivity, would you like to add this again?

The0x539:
wait, why

PR #687, opened 2015-11-30
Fixed the bottomsheet library version because android studio says so

rosenpin:
Changed to specific version


l3d00m:
is it ok when i push this in my next commit to github?

rosenpin:
It's ok

okungnyo:
Some info that might help settle you guys' checkbox vs. switch tug-of-war:

https://stackoverflow.com/questions/21231345/when-to-use-checkbox-and-when-switch-android
https://ux.stackexchange.com/questions/63884/checkbox-vs-toggle

Basically, the argument is strong on both sides lol.

l3d00m:
I think we have mostly the on/off situation, so it is better to use switches. Thanks for the links.

PR #688, opened 2015-12-01
Properly colorize status bar in comment settings

recalculated:


ccrama:
Good find, thank you!

PR #695, opened 2015-12-02
Move color theming to BaseActivity

recalculated:
Almost all classes that extend BaseActivity apply some sort of color theming to the activity background/toolbar/status bar, so it doesn't make sense to implement theming in each child independently when the statements are almost exact copies of each other. Doing that not only makes for more typing, it causes widespread inconsistency (for example, the Profile activity getting the status bar color with `Palette.getDarkerColor(Palette.getDarkerColor(Palette.getDefaultColor()))` while every other activity uses `Palette.getDarkerColor(Palette.getDefaultColor())`) and makes the code hard to maintain.


ccrama:
Wow, really great PR. Thank you very much for all the hard work you do!

PR #696, opened 2015-12-02
Add refresh button to action bar menu

recalculated:
Adds an action bar menu item to refresh instead of having to scroll all the way to the top to swipe to refresh.


okungnyo:
YES!!

ccrama:
I love you. Thank you!

ccrama:
This seems to remove shadowbox from the menu

ccrama:
Setting showAsAction to always fixed it, but now looks a bit jumbled. Wonder if there's a better place for Shadowbox mode or one of the other buttons...

alexsmithfanning:
Overflow menu? Refresh should go there anyway, shouldn't it?

ccrama:
Overflow would still add another button, still just as jumbled.

okungnyo:
The Shadowbox mode is not really an 'action' right?

Can't it be moved to the navigation drawer, then?

ccrama:
It's an action that only applies to the current page, just like refresh and sorting and info

okungnyo:
What do you mean by "jumbled"? Can you show us a screenshot?

Also, side note: would you consider renaming it to [Shadowbox View]? That fits better contextually.

ccrama:
I actually just removed the info button as it was mostly useless, looks better now :). And everything is branded with Shadowbox Mode right now, would take some work to change all of it

okungnyo:
That's too bad, 'Mode' implies that it's a transformation of the entire app, while 'View' only the current page, like you described.

ccrama:
Well it does open a new activity and transforms the current dataset, so mode seems more appropriate

alexsmithfanning:
Put refresh and filter into the overflow menu. Now there's only two buttons. No jumble.

ccrama:
3 buttons is fine, removing Info worked (no need to over complicate it with an overflow)

recalculated:
Can you upload a screenshot? When I'm on the frontpage, refresh, sort, and shadowbox are all showing, and when I swipe to a subreddit, info and shadowbox are moved into an overflow menu

ccrama:
Frontpage automatically hides info, which means the rest of the menu items expand out. If you have a different sub set as default, it won't look right

recalculated:
Was commit 6d9055d meant to remove the info button?

ccrama:
Yes, just realized that I messed up haha. Will fix in an commit later tonight!

recalculated:
Ah, ok.

theRealPadster:
This hasn't been added yet has it? I was thinking it would be a good idea and was going to suggest it when I found this.

Alexendoo:
It has been added, however the change may not have hit the play store version yet

theRealPadster:
Okay, cool. Thanks.

PR #700, opened 2015-12-04
Remove hardcoded strings, use showAsAction=ifRoom

recalculated:
Menu items generally shouldn't use `showAsAction=always` because it causes more icons than there is space for to appear on smaller devices.


ccrama:
Yeah my fault.Thank you!

Issue #691, opened 2015-12-01
Crash when spoiler link opened


silentkun:
Subreddits that use spoiler tags creates a link to a nonexistent page in reddit. This causes the app to crash when opened. 

[Example] (http://reddit.com/r/anime/comments/3uybmz/what_are_some_issues_in_an_anime_that_could_have/cxircne?context=3)

Alexendoo:
Related #692

l3d00m:
This is fixed now,  it no longer crashes

PR #705, opened 2015-12-06
Add spoiler functionality

### LINKED ISSUE #692: Support for spoilers
Subreddits such as /r/anime uses spoiler tags which do no show inside the comment.

Deadleg:
This took longer than I thought... I'm fairly confident it works, but I wouldn't be surprised if it breaks in places since spoilers are consistent between subreddits.

+++ COMMIT MESSAGE +++

This has been tested on subreddits that use
- #s
- /s
- /sp
- /spoiler

as spoiler identifiers. The effect is that the visible text is
displayed as a link, with a whitespace, then the spoiler text
blacked out. When the visible text the clicked, the text of the
spoiler text is reset to white, with the backgroud to stay black.

This is a bit fragile due to the fact that:
1. subreddits use different hrefs for spoiler tags,
2. the visible spoiler text is different per subreddit (and even
   per post).

Accordingly, there are a few assumptions. Firstly:

A1. Spoilers are identified by the presence of the "title" attribute
    within the <a> tag.

This seems consitent with the html that comes out of reddit, and
is used in the regex to find each spoiler.

A2. The text within the spoiler does not end in "]]".

The whole parsing token is "]s]]", but this is relaxed to just "]]"
since it may or may not happen. Should probably be strengthed in
the future.

A3. If the title attribue is empty, insert "spoiler" as the visible
    text.

This is fairly self-explanatory, since having no title will mean
there is nothing to click on.

If more patterns need to be added to identify spoilers, see
ImageType.getImageType(String url).

To prevent collapsing of the children tree when a spoiler link is pressed,
ActiveTextView.spoilerClicked is set to true and is checked in the onClick
function in CommentAdapter for the R.id.content element.

Some amount of code clean up was added, since there were a lot of
inconsitent spacing issues and a few indents which could be reduced
by re-wording if statements.

Fixes #692, #691

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/ActiveTextView.java
+++ b/app/src/main/java/me/ccrama/redditslide/ActiveTextView.java
@@ -44,6 +44,7 @@ public class ActiveTextView extends TextView {
 
     private boolean mLinkSet;
     private String mUrl;
+    private boolean spoilerClicked = false;
     private SpannableStringBuilder mSpannable;
     private ActiveTextView.OnLinkClickedListener mListener;
     private ActiveTextView.OnLongPressedLinkListener mLongPressedLinkListener;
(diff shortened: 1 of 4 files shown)


ccrama:
This is really great work, thank you very much! 

I was putting this off because of how complicated it is, can't wait to test it out!

Really grateful to have such great people helping me out with Slide :)

okungnyo:
Fan-f'ing-tastic!

PR #717, opened 2015-12-10
Added reddit settings

rosenpin:
The settings open the https://www.reddit.com/prefs/ link in the users browser to configure general reddit settings.
Example


ccrama:
Thank you so much! This is so awesome :)

PR #718, opened 2015-12-10
Fix crash loading app with disabled notifications

### LINKED ISSUE #598: Clicking 'Done' in notification settings causes app to crash [replication steps]
1. Disable notifications in the app settings
2. Swipe the app out of the recents screen
3. Enable notifications in the app settings and click done

(better replication steps thanks to @recalculated)

Deadleg:
When the app is started with the user having no notifications,
Reddit.notifications is null because the code that would initialize
it checks if Reddit.notificationTime (== -1) != -1.

All we have to do is check if it is not null, and instantiate it if
it is.

Fixes #598.


l3d00m:
Really good catch, thank you for your help!

PR #722, opened 2015-12-12
Fixed a stupid bug

rosenpin:


l3d00m:
This fixes that the reddit icon was too small, right?

rosenpin:
Now it does.

l3d00m:
Thanks for that!

PR #724, opened 2015-12-12
multiple fixes

l3d00m:
-extend all activities from baseactivity
 
-hide reply button on archived posts too
-Show a select dialog for opening a reddit URL outside of slide instead of hardcoding chrome
-some unfinished activities fixes

I don't know if the extending thing is the recommended way of doing this, that's why I'm making a pr.


ccrama:
Awesome, thank you!

l3d00m:
<3

PR #725, opened 2015-12-12
Fix spoiler link not working

Deadleg:
Was grabbing the start of the spoiler link, not the end. Must have done a last minute change but didn't test it.

#692


ccrama:
Ah, thank you!

PR #742, opened 2016-01-04
Removed \n string from Album View (Imgur)

KonradIT:
As seen in the Imgur Album view, when there is a GIF it says TAP TO LOAD GIF, In a new line, and the \n is not parsed to make a new line.

https://imgur.com/a/F1dd9

I removed the \n ~~(sorry, no after fix screenshot because my android studio is crapping on me)~~

EDIT: updated AS to 1.5.1, here is pic: http://i.imgur.com/NWsXgbA.jpg


KonradIT:
someone merge this pls

Alexendoo:
The navigation bar shouldn't be tinted, see #618

KonradIT:
Well, fixed! I reset --hard to the commit of removing \n

Sorry for the disturbance

ccrama:
Good find, thanks!

PR #750, opened 2016-01-10
More elegant way of setting view visibility

thatJavaNerd:


ccrama:
Good commit! Thank you

Issue #751, opened 2016-01-10
Issue saving Reddit preferences


ccrama:
Stacktrace:

```java
java.lang.RuntimeException: An error occured while executing doInBackground()
	at android.os.AsyncTask$3.done(AsyncTask.java:304)
	at java.util.concurrent.FutureTask.finishCompletion(FutureTask.java:355)
	at java.util.concurrent.FutureTask.setException(FutureTask.java:222)
	at java.util.concurrent.FutureTask.run(FutureTask.java:242)
	at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:231)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587)
	at java.lang.Thread.run(Thread.java:818)
Caused by: java.lang.IllegalArgumentException: Could not find value for ["en","en-us"]
	at net.dean.jraw.AccountPreferencesEditor.val(AccountPreferencesEditor.java:52)
	at net.dean.jraw.AccountPreferencesEditor.<init>(AccountPreferencesEditor.java:35)
	at me.ccrama.redditslide.Activities.SettingsReddit$1.doInBackground(SettingsReddit.java:41)
	at me.ccrama.redditslide.Activities.SettingsReddit$1.doInBackground(SettingsReddit.java:38)
	at android.os.AsyncTask$2.call(AsyncTask.java:292)
	at java.util.concurrent.FutureTask.run(FutureTask.java:237)
	... 4 more
```

Appears to be an issue with JRAW. Any idea of why this would be occuring @thatJavaNerd?

thatJavaNerd:
In that account's [preferences](https://www.reddit.com/prefs), what is the interface language?

ccrama:
@thatJavaNerd probably not english, but I never override the language when retrieving or saving the AccountPreferencesEditor, though.

thatJavaNerd:
This specific error occurs when [`AccountPreferencesEditor.val()`](https://github.com/thatJavaNerd/JRAW/blob/master/src/main/java/net/dean/jraw/AccountPreferencesEditor.java#L40-L53) tries to identify the type of object a given JsonNode is. It knows how to handle a null object, boolean, int, String, and an array of one element. The value for `lang` is an array of two elements so it doesn't know how to handle it and throws an error.

[...]

PR #754, opened 2016-01-10
Reference static variable through a static context

thatJavaNerd:
Static variables should not be referenced in a non-static context.


recalculated:
Why is a public, static field called `mContext` anyways? The `m` prefix is for non-public, non-static fields... It should be called `sContext` instead. Perhaps renaming it would avoid some confusion?

https://source.android.com/source/code-style.html#follow-field-naming-conventions

ccrama:
Good find! Also, do you mind looking at #751 ? Seems to be a jraw issue but it may just be my fault. Thank you for the commit!

okungnyo:
Has it been changed to sContext?

l3d00m:
@okungnyo yes

okungnyo:
In which commit?

PR #759, opened 2016-01-11
Moved startup_* strings to a <string-array>

thatJavaNerd:
Moving the constants into a `<string-array>` makes for less verbose code.

I'm not sure if the startup strings are still being used but if they are this will be a more succinct way of accessing them.


l3d00m:
A string array looks better, good find. They are still used on first startup and after first login. Thank you!

PR #768, opened 2016-01-12
Replace "Add friend" and "Remove friend" with XML constants

thatJavaNerd:


ccrama:
Thanks, should have done that. Also, would you be interested in joining us on IRC? If you are, the server is irc.snoonet.org and the channel is #slide!

thatJavaNerd:
I'll check it out!

ccrama:
Awesome!

PR #812, opened 2016-01-16
Fix spoiler text not revealing

### LINKED ISSUE #808: Currently need help with these features
There were a lot of requests from the thread on /r/android yesterday, here's the main things that I think need to be done and would like help on if you are looking for something to contribute to!

* [x] Spoilers are not working properly
* [x] Option to remove crop on comment view~
* [x] Option to color the navbar
* [x] Scrollbar in comments
* [x] Get rid of pinning system, create screen to re-order, remove, and add subreddits
* [x] Smaller image option for list mode
* [x] Loading comments with context doesn't scroll to the context comment (the highlighted one)

Also, here are a list of the biggest crash issues currently

http://pastebin.com/dPDeFP0q

If you would like to tackle something/have done something in a commit, please comment below!

Deadleg:
Commit e6accd550e6615d3b7c5c099030f042209fd385d changed the endOfLink
arg of setOrRemoveSpans to be the length of url, not the end position
of the text displayed.

Helps fix #808.

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Views/MakeTextviewClickable.java
+++ b/app/src/main/java/me/ccrama/redditslide/Views/MakeTextviewClickable.java
@@ -339,7 +339,7 @@ public class MakeTextviewClickable {
                     if (!clickHandled) {
                         // regular click
                         if (link.length != 0) {
-                            onLinkClick(link[0].getURL());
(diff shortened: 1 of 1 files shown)


ccrama:
Ah, thank you very much! Was wondering what I messed up :(

Issue #800, opened 2016-01-15
Loading posts without [Always show offline data first] still takes twice as long as other Reddit apps


okungnyo:
Of course, I turned off [Always show offline data first] for fairness.

I used Relay for Reddit for comparison:

https://play.google.com/store/apps/details?id=free.reddit.news
https://play.google.com/store/apps/details?id=reddit.news

And the difference in loading speed was like 1-2 seconds.

Although that doesn't sound like much, if you visit multiple subs, those seconds really do add up and make a huge difference.

I imagine speed is an important factor for many other users as well.

Alexendoo:
I can't reproduce this on my end, if you're swiping through subreddits a lot at a time to get to the one you want you may be hitting some rate limiting by reddit is a possible explanation of what you're experiencing

okungnyo:
Kill both apps, then load up Relay and see how long it takes to load your frontpage. Do the same with Slide. The difference is there, I assure you.

ccrama:
I did testing and it was near identical in speed. What kind of network are you on? Slide also preloads images so that might contribute

ccrama:
Also, Relay uses cookie authentication which is less secure (and might have to be changed due to Reddit's new usage TOS) but is faster.

okungnyo:
Ah, so that's why Relay is faster.

Is there really other no way to improve speed without making Slide insecure? That's a bummer...

ccrama:
Yep it is

okungnyo:
Two related questions.

1. On a cold start, does Slide load just the frontpage? Or does it predict and preload other subs too?

2. And does Slide preload any pics or GIFs?

ccrama:
1) Just loads the visible sub and the one(s) next to it.

2) Preloads pics

Issue #753, opened 2016-01-10
Shadowbox viewer only loads however many posts have loaded


parkerlreed:
If I jump to a subreddit and hit the gallery view button (shadowbox I think it was called?) it loads and lets me swipe through images but only for the first page of results. I have to back out and scroll down to load more posts. Then when going back it's reset to the first image, so I have to swipe about 20 times just to get back to where I was.

I guess what might work here is load more posts when you hit the last image in the viewer and also remember position if you happen to go back in.

Alexendoo:
Same occurs with swiping through posts - #478

l3d00m:
Marking this high priority because it makes one pro feature nearly unusable. It was also multiple times reported in other locations.

l3d00m:
fixed with bd7251d0c8712e1aba0a4b69f0b4a669df6e5ff4

PR #818, opened 2016-01-16
Refactor submission loading + load more posts when swiping

### LINKED ISSUE #478: When swiping through posts, there is a dead end and no more load
When swiping through posts (say, on the front-page where I know they're infinite) I will often get to a dead end after about 20-50 (haven't counted) to what I presume is the last post cached.

Somehow the next posts need to be preloaded to fix this and allow me to continue swiping.

This can be sometimes temporarily solved by backing out to the sub, then going back into the last post I saw

Deadleg:
This was more involved than I though it would be, but I think the end result is less complicated than the diff would show.

Essentially, SubredditPosts.LoadData#onPostExecute was refactored where view specific logic was moved to the SubmissionsDisplay interface. The fragments that need to be updated when more comments are loaded can implement their logic using this interface, which leaves SubredditPosts view agnostic. This allows fragments such as CommentsScreen and SubmissionView to both utilize this class, despite having different requirements.

In addition, when swiping through posts the comments async task is cancelled. Otherwise if you swipe rapidly, and a lot of the tasks are executed and you have to wait a long time for all the previous posts to be loaded.

There was a fair amount of code clean up too, but original logic should be intact.

I probably don't have time for this, but the next step I was going to take was make SubredditPost/MultiredditPost have private constructors, and use a factory to manage each subreddit. This would remove the need for DataShare.sharedSubreddits as its nasty.

Fixes #478, and allows #753 to be fixed pretty easily. Mildly related to #800


ccrama:
Very impressive work, thank you so much! I agree, the way I handled SubredditPosts and the shared dataset are quite ugly, and redoing is something I've been thinking about for a while. Also, good thinking about cancelling the comment loading task! 

This is definitely the correct way to implement loading more posts, and it works great. Just want to thank you for the high quality commits you've been contributing to Slide, and you're truly making it a better app :).

If you don't have time to work on the SubredditPost/MultiredditPost constructors and factory, let me know what your plans were and I will look into implementing it!

Deadleg:
Basically SubredditPosts would have a static method `public SubredditPosts getSubreddit(String subreddit)` that can be called to get an existing subreddit, or load posts for the first time. Internally, SubredditPosts has a private `Map<String, SubredditPosts>` of subreddit names and their corresponding object. This way, there is at most once instance of SubredditPosts per subreddit, which allows caching of all previously fetched posts without faffing about with static lists.

ccrama:
@Deadleg I like the idea! Will definitely look into doing this. Also, my latest commit rewrites Shadowbox mode to work properly with your new interface! Really slick, thanks again!

PR #821, opened 2016-01-18
Independent Font sizing

530farm:
Independent font sizing working


ccrama:
Good work! Thank you :D

PR #823, opened 2016-01-18
Fix a typo that caused the comment view to crash when the small font …

joytrekker:
…setting is selected.


l3d00m:
Nice spot, thank you!

PR #839, opened 2016-01-20
Drawer: Fix flash of count icon on start

aaronpoweruser:
Hide the count badge by defualt. Show it after updating inbox count

Signed-off-by: aaronpoweuser <atlrudeboi1@gmail.com>


ccrama:
Ah, good one! Thank you :)

Issue #868, opened 2016-01-22
Don't treat image/album view background like a piece of paper now that it's translucent


The0x539:
The background shouldn't cast a shadow. Perhaps it should fade out the further to the right the image is swiped.

Nath5:
I like the idea of a fade out, not sure the shadow is wrong though, even a translucent sheet will cast some shadow.

The0x539:
It definitely *looks* wrong as it is right now. If it's gonna cast a
shadow, it should be underneath it too.

ccrama:
Better animation has been added (less paper-y)

The0x539:
http://i.imgur.com/Y2tJAqs.png

ccrama:
I am closing this issue as there is an option to disable the transparent background, and the enter animation has been adjusted accordingly (and the exit animation if you tap to exit)

PR #897, opened 2016-01-26
Add option for solid image viewer background.  Delay showing progress…

Nath5:
… bar on image viewer for 500 ms to stop it from flashing on quickly loading images.

TODO:
I don't know where you are pulling the icons from but the new settings will need an appropriate icon.

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Activities/Album.java
+++ b/app/src/main/java/me/ccrama/redditslide/Activities/Album.java
@@ -15,9 +15,11 @@ import android.view.MenuItem;
 import android.view.View;
 
 import com.afollestad.materialdialogs.AlertDialogWrapper;
+
 import com.google.gson.JsonArray;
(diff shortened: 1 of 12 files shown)


The0x539:
Hmm, this does complicate #868 slightly.

Nath5:
I have updated the affected files using the ASOP style sheet found here:
https://github.com/android/platform_development/blob/master/ide/intellij/codestyles/AndroidStyle.xml

Please let me know if there is another style sheet I should be using.  Before I was using the default Android Studio styles.

Deadleg:
Looks better now, thanks! I'm using the default in android studio too but I don't get that kind of formatting. Strange...

ccrama:
Awesome! And yeah for future reference, you can click CTRL + SHIFT + L to auto-format code to Android standards using Android Studio.

Looks great, will edit the icon for you using one of the current icons in the resources directory! Thank you!

PR #898, opened 2016-01-26
Should fix the double last updated toast.

Nath5:
I think called setCurrentItem twice is what is causing bug 887 to occur.


ccrama:
Hey, thank you for the commit! Unfortunately, this is necessary to color the drawer views correctly. We'll have to find a different solution

PR #916, opened 2016-01-28
Made method for Inbox count badge update and added to onResume()

530farm:
Made method to replace the Async tasks that occurred a couple times in Main Activity, and added it to onResume() to update the badge count with out having to restart the MainActivity or going into the inbox itself.


ccrama:
Makes more sense to do it this way. Thank you very much!

PR #920, opened 2016-01-29
Refactor code

l3d00m:
Making a pr for this because this could possible destroy some functions. I watched closely while refactoring it, but I would prefer when another one can take a look at this, especially at the String constant refactoring.

**Changes:**
- Replace intent extra strings with string constants
- Replace shared preferences strings with string constants
- Small lint cleanup
- Replace static "Slide" log tag with dynamic log tags depending on the file and the linenumber
- Unify all setting values and their logic into SettingValues.java
- Remove some code duplicates
- Fixed bugs caused through different strings in sharedPrefs getter & sette


ccrama:
Very great changes! I especially like the fact that settings now live in SettingValues, that was something we should have done a long time ago. Will be testing it out and making sure it works, but it appears that it does!

Thank you :)

PR #925, opened 2016-01-29
Fix multiple code blocks not rendering

### LINKED ISSUE #921: Markdown: Code formatting doesn't work with more than one codeblock
If there is more than one code block, the submission/comment text is cut before the second code block. See https://plus.google.com/+PahoLurieGregg/posts/jMc4YULvJeR for an example.

Paging @Deadleg for help.

Deadleg:
Need to limit the split method to only split of the first occurance.
It is possible the rewrite this section no not do this split (which
would probably be more efficient), but this section does not incur
a large performance penalty as of yet.

Fixes #921


l3d00m:
This looks really good, especially the test classes. I really like your code style!

ccrama:
Great find. And I love your style too, glad to have you as a contributor! Thank you for all the work you've been doing :)

PR #963, opened 2016-02-06
Fixed a crash on the home screen when not logged in.

Gilleland:
Fixed a crash on the home screen when not logged in.


ccrama:
Thank you for the find and fix!

ccrama:
Wait sorry, had to revert this. What are the other four commits for?

Gilleland:
Merging changes from ccrama/master into the master branch of the forked repo I made, the only change you'll be merging in is that one line fix.  I normally use sourcetree and wouldn't leave those in.  Git knows what's up and knows those are already part of the current branch.

PR #971, opened 2016-02-10
Fix non-card submission thumbnails not being square

Plastix:
Here is a side by side comparison: 
card_demo

This PR fixes the thumbnail on the top.

### DIFF
--- a/app/src/main/res/layout-v21/submission_list.xml
+++ b/app/src/main/res/layout-v21/submission_list.xml
@@ -206,8 +206,7 @@
                     <com.makeramen.roundedimageview.RoundedImageView
                         android:id="@+id/thumbimage2"
                         android:layout_width="70dp"
-                        android:layout_height="wrap_content"
-                        android:maxHeight="50dp"
+                        android:layout_height="70dp"
                         android:layout_alignParentRight="true"
                         android:layout_centerVertical="true"
                         android:layout_marginRight="8dp"
--- a/app/src/main/res/layout/submission_list.xml
+++ b/app/src/main/res/layout/submission_list.xml
@@ -206,8 +206,7 @@
                     <com.makeramen.roundedimageview.RoundedImageView
                         android:id="@+id/thumbimage2"
                         android:layout_width="70dp"
-                        android:layout_height="wrap_content"
-                        android:maxHeight="50dp"
+                        android:layout_height="70dp"
                         android:layout_alignParentRight="true"
                         android:layout_centerVertical="true"
                         android:layout_marginRight="8dp"


ccrama:
Thank you for the PR! I have actually totally changed the layout in my new_offline_model branch (and subsequently fixed this issue), so this PR will get overridden when I merge, unfortunately. Feel free to PM me or make an issue if you would like to PR in the future, and I'll let you know if it's being worked on or not!

Plastix:
No problem at all. Thanks for taking the time to look at this. I just took a quick look at your new_offline_model branch and the layout has indeed changed dramatically. Great job at making the layout a lot flatter!

In doing this PR I've noticed that you are overriding layouts in `layout-v21`. This leads to duplicate layouts that are a pain to manage. You should consider switching to themes and overriding themes instead of layouts. This way you will only have one version of a layout.

ccrama:
Yeah, it has sped up scrolling a lot! I'm happy with it. As for the v21 layouts, that is to make use of the Boundless background ripple effect in lollipop and above, but I do agree it's a pain. I might extract the actionbar into its own layout, and then just include the actionbar as a <merge> into a single layout, which would make it easier to manage. Is that what you had in mind?

Plastix:
That is still a good idea but not what I'm talking about. The idea is to override styles instead of overriding layouts. That way you have multiple styles that you apply to a single layout.

For example: The reason why we need to override the layout is because of the ripple background attribute. Lets make some styles:

In `style.xml`

```XML
    <style name="Submission"/>

    <style name="Submission.List">
        <item name=">android:background=">android:selectableItemBackground</item>
    </style>
```

In `values-v21/style.xml`

```XML
    <style name="Submission.List">
        <item name="android:background">?android:selectableItemBackgroundBorderless</item>
    </style>
```

Then in `submission_list.xml` all you need is:

```XML
  <LinearLayout
                        android:id="@+id/base2"
                        android:layout_width="match_parent"
                        android:layout_height="wrap_content"
                        style="@style/Submission.List"
                        android:maxHeight="82dp"

                        android:orientation="horizontal"
                        android:paddingBottom="4dp"
                        android:paddingLeft="16dp"
                        android:paddingRight="16dp"

                        android:paddingTop="8dp">
...
```
So in this case you only need one layout and the style is changed depending on the system version. I haven't tested this but I'm pretty sure it should work. Since you are refactoring layouts you might as well try this. It would definitely make your layout code easier to manage.

ccrama:
Ah that's great. Just changed it on all the views and will commit it in a minute! Thank you for the heads up, much better solution to this issue.

Plastix:
Here are a few articles on Android styles that you might want to check out. 

- http://blog.danlew.net/2014/11/19/styles-on-android/
- http://mdswanson.com/blog/2013/12/11/responsive-layouts-in-android.html

PR #977, opened 2016-02-13
Update to imgur api v3

### LINKED ISSUE #867: Galleries do not load more than 10 images
How do I add the "BUG" label?

Attached is a link to a Reddit post containing a 63 image album.

Slide will only show the 1st 10 images both in shadowbox or in regular view, with no option to load the rest.

Latest beta.

Albums not completely loading

https://www.reddit.com/r/pics/comments/41aexc/this_guy_found_a_huge_underground_facility/

Google+ bug report: https://plus.google.com/+ZalmanEichorn/posts/LLKgaW1yMZ8

l3d00m:
#### TODO:

- Help wanted: **Use new imgur api for gifs** (instead of gfycat because it doesn't work any longer) 

- Waiting for reply from imgur (concerning heavy usuage)

- Needs testing because I just rebased from master(02/18/16)

#### Description:
The new imgur api makes it easier to parse the answers, which allows it to make the code more clean.
Besides that, code in AlbumUtils.java was refactored because of unneeded asynctasks

this also fixes #867 

#### Known bugs:
* Load album is not displayed anymore and instead freezes the app

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Activities/Album.java
+++ b/app/src/main/java/me/ccrama/redditslide/Activities/Album.java
@@ -56,7 +56,7 @@ public class Album extends FullScreenActivity {
 
         String url = getIntent().getExtras().getString(EXTRA_URL, "");
 
-        new AlbumUtils.LoadAlbumFromUrl(url, this, true, getSupportActionBar(), (RecyclerView)findViewById(R.id.images)).execute(url);
+        AlbumUtils.loadGalleryAndDisplay(url, this, true, getSupportActionBar(), (RecyclerView) findViewById(R.id.images));
 
     }
 
--- a/app/src/main/java/me/ccrama/redditslide/Activities/Album.java
+++ b/app/src/main/java/me/ccrama/redditslide/Activities/Album.java
@@ -56,7 +56,7 @@ public class Album extends FullScreenActivity {
 
         String url = getIntent().getExtras().getString(EXTRA_URL, "");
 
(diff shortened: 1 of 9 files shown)


ccrama:
Can't wait to merge this, just waiting for a reply from Imgur about whitelisting Slide :)

ccrama:
It appears imgur changed their stance on free applications, and it appears you no longer have to get whitelist status to use the API for free. From the website,

"Free Usage
The Imgur API is free for non-commercial usage. Your application is probably free if you don't plan on making any money with it, or if it's open source."

 This would also explain why I haven't gotten a response from them about whitelisting slide. I think we're good to merge

Alexendoo:
That was always the case, we were applying to not be restricted by their rate limiting since 12,500/day may not enough, the limit though is still apparently in place

ccrama:
They changed the section about whitelisting and removed the whitelist page, and put that blurb in there instead. I believe you only get restricted now if it's for commercial usage, which you have to pay for (the paragraph before)

l3d00m:
Imgur doesn't allow exceeding the rate limit anymore for free apps. We will use the old api for now.  
Note to self: Their support is very unresponsive.

okungnyo:
What advantages does the v3 API have over the old free API?

Why would anyone use it when they can just continue to use the free one?

I'm confused.

Alexendoo:
Because the old one was at that point long deprecated, and I believe no longer exists?

l3d00m:
Originally the v3 was free for open source applications. They removed the free acces that shortly after we requested it. 

This whole thing was also before they closed down the old v2 api.

okungnyo:
All right, thanks.

I guess they had to find some way to monetize their service.

Issue #968, opened 2016-02-09
Spoilers in titles are not blacked out


Deadleg:
I think this is pretty trivial to fix with SpoilerRobotoTextView.setTextHtml in my branch, but I'll leave this here just in case.

Deadleg:
Turns out this done by css, so can't actually implement this (easily). Can be closed.

Issue #934, opened 2016-01-30
Markdown: Tables are not implemented 


alexsmithfanning:
Tables are incorrectly formatted. They don't include sections or anything, just a line of text.

**A table in Slide:**
![screenshot_2016-01-29-19-59-51](https://cloud.githubusercontent.com/assets/12200557/12692932/aee20c7c-c6c4-11e5-8e6d-93a867b1cfb0.png)

**A table in normal Reddit:**
![reddit-tables](https://cloud.githubusercontent.com/assets/12200557/12692953/e0ca8af2-c6c4-11e5-8d99-ce6009e96c4e.png)

Issue #682, opened 2015-11-28
Markdown: Nested Lists


karlding:
Nested lists fail to display the proper nesting levels.

l3d00m:
this is kinda related to #40

youri--:
I've just tried nested lists in a comment and neither the preview or the comment rendered the nested list properly.

Nested unordered lists need 2 or more spaces at the start of a line, ordered lists 3 or more spaces, according to https://www.reddit.com/wiki/markdown#wiki_lists

Can this issue be reopened please?

l3d00m:
@youri-- I'd just recommend to open a new one, it'll probably get more visibility than an old issue like this.

Issue #440, opened 2015-11-08
Markdown: Ordered Lists


karlding:
Ordered lists are being treated as unordered lists.

```
1. Item 1
2. Item 2
3. Item 3
```

Should display as

1. Item 1
2. Item 2
3. Item 3

Not

* Item 1
* Item 2
* Item 3

TomerRosenfeld:
Can you please provide a screenshot example?

karlding:
This is because of the replacements that ``ParseTextWithLinksTextViewComment`` does on line 116 in [MakeTextviewClickable.java](https://github.com/ccrama/Slide/blob/06dec494eb9a2f7c6199805498209c72601108d8/app/src/main/java/me/ccrama/redditslide/Views/MakeTextviewClickable.java#L116). An ordered list is preceded by ``<ol>`` and then each list ``<li>`` element, but currently, the code blindly replaces any ``<li><p>`` with ``<p>• ``, assuming that the only lists are unordered lists.

The problem is that it's being treated exactly the same way as an unordered list.

PR #978, opened 2016-02-13
Better text view

Deadleg:
(Same as previous pull but merge into master)

Almost finished

I've broken a fair amount of stuff and I need help in getting it all fixed to how it was. I think most of the work will be refactoring the views, and fix possible issues where MakeTextviewClickable was used. Maybe it would be good if this were merged into a separate branch?

This started off as a pull request to display tables, but this has since turned into much more of a refactor due to performance issues.

It looks like the scrolling performance of every list view can be fixed to reducing the nested Linear/Relative layouts in the xml. The diff in comment.xml is an example of this flattening, and it's really been the biggest factor in speeding up scrolling.

MakeTextviewClickable has also been removed, since a new instance is created on every onBindViewHolder call which is really unnecessary and had some funny logic. In doing so I've ripped out and bunch of code of with the old references and replaced it with new code that may or may not work. The most complete example is in CommentAdapter.setViews.

Current features are
* horizontal scrolling code blocks
* tables
* strikethrough
* ordered lists
* nested lists
* increased comment scrolling performance

I aim to solve the following (at least) with this pull:
* #934 done
* #440 done
* #682 done
* #122 done
* #968
* comment scrolling performance (mostly there)


ccrama:
Very cool! Can't wait

l3d00m:
The hype is real! This is really great

ccrama:
Woo! Finally merging. Looks like everything is good and working!

PR #1003, opened 2016-02-16
Hide post only for signed in users

### LINKED ISSUE #989: Hide post should not display for signed out users
Currently the option to hide a post is available to users who are not signed in, and it temporarily hides posts and gives the hidden forever message, however an account is required for this feature so on refresh the post returns

rigardin:
Fixes #989


ccrama:
Thanks for the fix! There are two other places hide is implemented, but I will replace them with your line :)

rigardin:
Sounds good, thank you!

PR #1015, opened 2016-02-17
Minor text fixes + remove MakeTextviewClickable

Deadleg:
Looks like my last commits broke tables, but they're fixed now and with a bit of padding between columns.

Text size now scales with users preference in tables and codeblocks + everything after.


ccrama:
Very great work! Thank you!

PR #1022, opened 2016-02-18
Fix issue with multireddits no loading

Deadleg:
Reproducing the error is a bit hard, I was on data at the time and I could load my multireddits as described below.

There is one assumption in that if something wants multireddits, then they will want to load them if they have not been already.

Commit message:

Slide could get into a state where the SubredditStorage.multireddits
field would be empty due to a network error/some other error.

Attempting to view multireddits would result in a error dialog due to
a NullPointerException due to SubredditStorage.multireddits being null,
which would return you to the main activity. Attempting to view
multireddits again would lead to the exact same message, since there
is no attempt to reload the multireddits in this action.

Adding a getter for multireddis gives an opportunity to load
multireddits if it is null, which should only be if it failed in
the past.


ccrama:
Great find! Thank you!

PR #1041, opened 2016-02-21
Fix horizontal scrolling in sidebar

### LINKED ISSUE #1036: Scrollable items in the sidebar can only be scrolled to the right
For instance see [/r/bestof](https://reddit.com/r/bestof), the sidebar has some pretty wide tables, however you can't scroll to the left as it closes the sidebar

edit: Seems you can scroll it to the left again by initially scrolling right, then without releasing scrolling to the left, but this is not desirable

Deadleg:
The DrawerLayout captures the close swipe action before it is propagated to its children. Overriding DrawerLayout.onInterceptTouchEvent allows the sidebar to check if the event coordinates hits a scrollable view, and stops the SidebarLayout from capturing the event if it does.

Fixes #1036.

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Activities/MainActivity.java
+++ b/app/src/main/java/me/ccrama/redditslide/Activities/MainActivity.java
@@ -48,6 +48,7 @@ import android.widget.ArrayAdapter;
 import android.widget.CheckBox;
 import android.widget.CompoundButton;
 import android.widget.EditText;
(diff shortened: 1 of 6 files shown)


ccrama:
Awesome, thank you! The same thing occurs with sub tabs disabled in comment views, will take a look at how you implemented it here and see if I can remedy it in comments as well :)

PR #1060, opened 2016-02-23
Cancel loading when swiping through multireddits

Deadleg:
Fixes an issue where if you swipe through a number of multireddits
it can take a while to load the submissions for the final multireddit,
since the submissions for all of the other multireddits are being
loaded first.


ccrama:
Oh dang, good find. Thank you for the fix!

PR #1088, opened 2016-02-26
Expand collapsed comments with a single press when (SettingValues.swap == true)

Nxt3:
Sorry for all of the commits--I love/hate git sometimes.


Deadleg:
If you do `git rebase -i HEAD~7` (7 is the number of commits) you can squash them all into one commit.

PR #1096, opened 2016-02-28
Single tap to expand a collapsed comment if (SettingValues.swap == true)

Nxt3:
When a comment is collapsed, the user can single tap on it to expand the comment if they have swap == true.


ccrama:
Woo, thank you for merging the commits, and thanks for the PR!

Nxt3:
Anytime :D

PR #1102, opened 2016-02-29
Add support for custom sub emotes to display inline

eyecreate:
Some subs, like /r/mylittlepony, have custom emotes they use inside comments and posts. An Android app was made to sync these large collections offline and some apps, like Relay for Reddit, read them to display in app. This just brings this functionality to Slide for Reddit.

p.s. at some point, SpoilerRobotoTextView should be renamed to match the fact it's the text renderer for most the app.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -85,10 +85,6 @@ dependencies {
     compile 'com.android.support:multidex:1.0.1'
     compile 'org.jsoup:jsoup:1.8.3'
     compile 'it.sephiroth.android.library.targettooltip:target-tooltip-library:+'
-    compile 'com.github.NightWhistler:HtmlSpanner:0.4'
-    //Resolves test and main app version conflict
-    compile 'org.apache.ant:ant-launcher:1.8.0'
-    compile 'org.apache.ant:ant:1.8.0'
 
     testCompile 'junit:junit:4.12'
     testCompile 'org.mockito:mockito-core:1.10.19'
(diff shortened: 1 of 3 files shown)


Deadleg:
The HtmlSpanner breaks a lot of the existing formatting since it does it's own thing with text formatting. See [this post](https://www.reddit.com/r/raerth/comments/cw70q/reddit_comment_formatting) as an example. 

It follows that node handlers need to be unregistered or move the existing code to use node handlers. Migrating the existing stuff should be fine except maybe tables, it's just a wee bit of work.

eyecreate:
Ah, you are right. It's too bad since the builtin Html parser doesn't allow you to override handling standard nodes.
I'm not feeling motivated enough to make all of them handlers, so I'll try and work it into a normal span iterator.

Alexendoo:
Hey can I ask what other subreddits use this type of thing? I've not not come across/noticed it on any I've been on

eyecreate:
There is a list here for MLP emotes:
https://ponymotes.net/bpm/
If any other subs besides those use this, I don't know.

ccrama:
Great work, and I'm glad you could do it without the extra dependencies and with a method that fit more with the code already present. And I do agree SpoilerRobotoTextView should be renamed, will look into that.

I would be really interested in the future extending this capability to support more subreddits with emoticons (if there are any non-MLP emotes used out there), will look around to see if any other subs have similar features.

Thank you for the PR!

ccrama:
Works, thank you!

Issue #1116, opened 2016-02-29
Open externally by domain


Alexendoo:
For an example see RiF, but it's a pretty nice addition, for instance if somebody has an app for one of the streamables that handles it better, or wants to use opengur etc. they would head to link handling and type the domain in

SZE2:
This would be great, as Imgur is still pretty slow on Slide vs. Opengur. Maybe see if you can implement their way of getting gifs? It's on github under opengur

Alexendoo:
Would also resolve #190

ccrama:
Added in in v4.7.4!

PR #1117, opened 2016-02-29
Open all youtube links in app

Alexendoo:
Reasoning:

1. The app itself is better than the mini youtube player given currently
 * It follows timestamps, this is a pretty big one
 * Better approach to fullscreen - tap once or go landscape
 * You can like/dislike videos
 * views, comments, description etc all visible
 * It supports 360/3D videos, often pop up on /r/videos etc
2. It won't crash for people on debug builds or without the youtube player installed
3. It removes a .jar from the project, we're open source and the source to it is not available, it may be ugly under the GPL

On top of this streamable support is technically still there for youtube videos, if in the future #1116 gets implemented, it could be run through the normal open method but having youtube domains pre-added to the list, giving the user the choice to open it in the streamable if they wish


ccrama:
Alright. I think this is the best course of action, especially since it breaks GPL. Thank you for the PR!

PR #1120, opened 2016-02-29
Allow user to add subreddits with spaces in them

Nxt3:
For example, when typing in "android dev", the Asynctask will run with the query "androiddev".


ccrama:
Great idea, thank you for the PR!

PR #1122, opened 2016-02-29
AutoFit text for multireddits displayed in drawer

Nxt3:
Not sure how you feel about adding another library in. Either way, the results are pleasing.

**Before**
screenshot_20160229-140722

**After**
screenshot_20160229-141126


Nxt3:
Library used: https://github.com/grantland/android-autofittextview

ccrama:
Actually yeah I really like that! How much overhead does the new library add (size of lib)?

Nxt3:
How exactly do I find that out?

PR #1140, opened 2016-03-04
Fixed some more spacing issues in Search's toolbar

Nxt3:
I started to change the layout of the messages in the Inbox--but I realized that the large marginLeft was because it's supposed to be a "reply" to your message. So nothing really changed in the _message_reply.xml_.


ccrama:
Good find, thanks! And yep the margin is for replies

PR #1170, opened 2016-03-09
Removed Internal Video handling settings

Nxt3:
Since all videos are handled in YouTube now--this setting does nothing


ccrama:
Good one, thanks! And sorry haven't gotten a chance to look at your nav bar PR because of the release and all, did you end up closing it?

Nxt3:
I closed it because it was becoming difficult to keep rebasing on all your settings commits. :P That, and I had this commit in there, which I shouldn't have done.

I'll work on it offline and submit a PR when I'm closer. The issue I was having was the contents of windows being messed up near the toolbar (i.e. toolbar going under statusbar). The attribute `fitsSystemWindows=true` needs to be set in certain layout files--but I don't know enough about how everything is intertwined to know which layouts to target.

PR #1211, opened 2016-03-20
Add history management to slide

tasgoon:
I created a new history settings page, where one can enable history tracking and whether to track seen NSFW posts. Keep in mind though, that loading the history settings page doesn't work quite correctly (the main settings activity becomes black as the history settings activity loads), but everything else works.


ccrama:
This is quite cool, great work! I will be merging this into my comment_swipe_layout branch for inclusion in v5. 

There appears to be some merge conflicts, but I'll work those out for you.

Thank you very much!

ccrama:
Done! Thank you for the work. I also made it work with the SubmissionAdapter, and updated the settings icon :)

PR #1248, opened 2016-03-25
Fixed NSFW videos and albums not being tagged properly

Nxt3:
Videos and albums weren't being tagged as NSFW


ccrama:
That was on purpose, NSFW shouldn't take precedence over videos and albums

Nxt3:
Why not? It works correctly

ccrama:
No, it shows as NSFW Image (there is no NSFW album or NSFW video tag)

Nxt3:
Oh. I could have sworn it showed for NSFW videos. My bad.

PR #1257, opened 2016-03-27
Correct offset for pull-to-refresh in SubmissionsView--hacky

Nxt3:
In my opinion, this seems hacky.

If you launch the app with a clean start--`getActivity().findViewById(R.id.header).getMeasuredHeight()` will return 0. However, once you're in the app and you navigate to another tab or single subreddit, then the height will be reported correctly until you enter a cold start again.

To curtail the returned 0, I set the `headerHeight` to default to `334`--after testing it on a Nexus 4, Pixel C, and Nexus 6P--this seems like a good middle ground to default to.

Furthermore, `mSwipeRefreshLayout.getProgressCircleDiameter()` will always return 0 no matter what; no idea why; instead, I just used `42` as the offset height from the `headerHeight`--which seems to look nice.

This is only for the `SubmissionsView`; I wanted to see what you had to say about it before carrying on with the remainder of the layouts.


ccrama:
Much better than the current implementation. Thank you!

PR #1259, opened 2016-03-27
Updated translations, fixed some hard coded strings

Soberat:
I really hope I won't crap anything up :D


Alexendoo:
Thanks for catching those ones, for the new translations though could you please submit them through https://crowdin.com/project/slide-for-reddit

Soberat:
Will do once I get home!

Alexendoo:
Thanks for that, from the way crowdin works it will override your changes since they're not in its own order, if you want to include translations themselves in your PR you can get the strings from https://crowdin.com/download/project/slide-for-reddit.zip, otherwise not a problem I'll do it at a later date

Soberat:
Done. By the way, would Crowdin be able to export plurals with "many" instead of "other"? It's a minor thing but in Android Studio it reports it as an error. Not a necessity, but would be nice :)

ccrama:
Hey just an FYI, that "donottranslate.xml" file should not be translated. Great work, though! Also feel free to join us on IRC

ccrama:
Thank you for all the hard work!

PR #1273, opened 2016-03-28
Collapse parent comments regardless of settings

Nxt3:
When the user is viewing comments and presses the "Collapse Comments" menu overflow item, it would take into account if the user had "Fully collapse comments" enabled. I find this useless when trying to collapse all comments, as the user is probably trying to hide all children comments.

This renames the overflow item action to "Collapse children comments" and does so regardless of the value of `SettingValues.collapseComments`.

In my testing--I couldn't find any issues with this approach. Everything seems to work properly. 

PS: I gave an incorrect description to my commit--I meant to say "children comments".


ccrama:
Great find, thank you!

PR #1279, opened 2016-03-29
Pull to refresh offset now starts behind toolbar in all views

Nxt3:
Now complete for all views! Works like a charm.


Nxt3:
Also, I removed unnecessary `findViewById()` calls in MainActivity when you are looking for the `R.id.header`. We already set `header` and I haven't had any issues with anything, so this is good. :P

ccrama:
Awesome, thanks!

PR #1280, opened 2016-03-29
Setting for collapsing comments by default

Nxt3:
Still have one minor issue:

If you go to your Inbox and click through to view a comment reply--once you're in that thread and you tap "Load full thread"--upon load, the comments won't be collapsed.


ccrama:
Great addition! Thanks. And I honestly think it's ok to have them all shown when you click "Load full thread", as it's inferring that you're loading the full thread

Nxt3:
No problem!

Good point. I actually know how to fix it--but then it causes the comments to be collapsed when viewing comments from the Inbox (Inbox -> Comment in question (these comments end up collapsed)). If you think it's fine, then I'll leave it as is.

PR #1286, opened 2016-03-30
Submission vote layout fix

Nxt3:
Flash will now be at the correct height


ccrama:
Great work, thank you!

PR #1293, opened 2016-03-31
IMPORTANT - Fix not being able to open comment menu

Nxt3:
If the user has `SettingValues.swap == true`, because of my last PR, it's now impossible to tap on a comment to open the comment menu.

I'm so sorry that happened. -_-


ccrama:
Yeah lots of people freaking out, will push immediately!

PR #1304, opened 2016-04-01
Fixed PTR offset in comments for np.reddit links

Nxt3:
If the comments were an np.reddit link, the PTR offset wouldn't be correct; also, if it was np.reddit AND had the "Load full thread" header, the PTR offset would sit partially behind the header.

If you want an example of the bugs this fixes, view these two links:

* np.reddit link: http://np.reddit.com/r/chemistry/comments/4cpbrh/almond_smell/
* np.reddit link & load full thread: https://np.reddit.com/r/chemistry/comments/4cpbrh/almond_smell/d1lebgw?context=3


ccrama:
Nice! Thank you for all of this work :)

Nxt3:
No problem! I enjoy it.

PR #1307, opened 2016-04-02
Remove minHeight from Toolbar instances in layouts

Gilleland:
This minHeight attribute was causing the toolbar to appear larger than necessary on some of my devices and causing the nav, up, and icons to appear above center.  My guess is this is used to serve some purpose but maybe isn't needed anymore?  If not, let me know.


ccrama:
Oh that's what's been causing it? Good find! No, they shouldn't have had it.

Thank you!

PR #1314, opened 2016-04-05
fix build error related to album_image.xml

septemus:
#no adverse side effects on marshmallow

if anybody on kitkat would like to test to make sure this works on KK before merge, I've uploaded an apk with the fix.

https://drive.google.com/file/d/0B-2xpDNqB2krNWNKU1VidGVERDQ/view?usp=sharing

apk is up to commit b76bc08 ("Fixed incorrect font color on light theme comment menu")


ccrama:
Works, thank you!

PR #1316, opened 2016-04-05
Collapse comments when long pressing on a highlighted comment

Nxt3:
If the user has `SettingValues.swap` enabled, and they've highlighted a comment (i.e. single click), and then they long press the comment--it should collapse, not just hide the menu.


ccrama:
Awesome, thank you!

PR #1325, opened 2016-04-07
Prevent a webpage to keep running (with sound) in the background when custom tab is closed

gferon:
Set FLAG_ACTIVITY_NO_HISTORY on the Chrome CustomTab Intent prior to launch. This prevents the webpage to keep running when the Activity is moved to the background.

Sorry for not submitting a bug report prior to making this PR, I really wanted to tackle this issue myself to get started :smile:

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/util/CustomTabUtil.java
+++ b/app/src/main/java/me/ccrama/redditslide/util/CustomTabUtil.java
@@ -57,7 +57,9 @@ public class CustomTabUtil {
                 CustomTabsIntent customTabsIntent = builder.build();
 
                 customTabsIntent.intent.setPackage(packageName);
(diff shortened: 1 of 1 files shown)


ccrama:
Great find, and thank you for the PR!

ccrama:
Actually, this had some adverse side effects (anyone that left the page could not go back to CCT), I have to undo this for now (unfortunately). Is there any other way to get the same desired effect without keeping it out of the backstack?

gferon:
I see. I don't really know a way yet, but I'll try to find one.

gferon:
Ok, I started investigating a bit and the issue does not occur with custom-tab-client demo (https://github.com/GoogleChrome/custom-tabs-client), I think I'll have to rewrite the way Chrome Custom Tabs are created to follow the best practices.

PR #1326, opened 2016-04-07
Fixes #1300 (Fixes refresh if no Settings changed)

### LINKED ISSUE #1300: Refresh even if no settings are changed
Slide version: 5.0.3
Android version: 6.0.1

Not really a bug but would help a lot if you would only refresh if the result of the settings activity was that anything has changed.

Nxt3:
Also, it now only refreshes the theme if a theme-related change was made. :)
#1300


ccrama:
Awesome stuff, thanks!

PR #1330, opened 2016-04-08
Use only top 500 safe for work subreddits in suggestions

Alexendoo:
Not sure if you want to do it this way or not hence the PR, but I think it makes sense personally


ccrama:
Yeah I was going to do this earlier, thank you!

PR #1332, opened 2016-04-08
Gradle Update

prashantkhurana:
Hey,
Android studio 2 was giving this popup " gradle 2 allows to use instant run". App runs fine with my little testing. Let me know if I am missing something.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -4,7 +4,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:1.5.0'
+        classpath 'com.android.tools.build:gradle:2.0.0'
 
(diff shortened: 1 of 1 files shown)


gferon:
Also debugged the app yesterday with these settings, works fine for me. Instant run is really cool apart from longer initial build times (because of instrumentation taking place ahead of time).

Alexendoo:
Thanks, and just as an aside you may or may not be aware of but it appears you haven't linked the email address you're using to commit with your github account

https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user

prashantkhurana:
Thanks @Alexendoo fixed.

PR #1343, opened 2016-04-08
Bold score text when voted

Nxt3:
When a submission is voted on, the score text will be bolded. This makes it easier to see on certain colored cards and makes it easier to tell if you've voted on a submission at a glance.


Nxt3:
Changing the order of the vote process to have it flash first makes the changing of the vote text Typeface change look smoother.

ccrama:
Great work, thanks!

PR #1349, opened 2016-04-09
Removed unused resources + externalise moderator strings

Alexendoo:


ccrama:
Awesome, thank you very much!

PR #1367, opened 2016-04-10
Multireddits embedding

prashantkhurana:
Allows user to import multi reddits with their name. The solution is to add "multi" to the name of the multireddit. I don't expect this to be merged yet, but wanted to see if anyone has any feedback. 

An elegant solution would be to redesign everything from arraylist of strings to arraylist of objects (each object has name, isMulti, description, sorting, etc) .But this is a simpler approach.

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Activities/MainActivity.java
+++ b/app/src/main/java/me/ccrama/redditslide/Activities/MainActivity.java
@@ -142,6 +142,7 @@ public class MainActivity extends BaseActivity {
     public boolean singleMode;
     public ToggleSwipeViewPager pager;
     public List<String> usedArray;
+    Map<String, String> multiNameToSubsMap;
     public DrawerLayout drawerLayout;
     public View hea;
     public EditText e;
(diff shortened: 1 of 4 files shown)


prashantkhurana:
@l3d00m @ccrama hey guys. I think this is done for it's initial phase. Here are the things that I know are pending :  
1) Not sure if this will be saved as part of google drive save. Didn't look into it.
2) Will work only for comment pane and no subreddit tabs. Doing it for tabs is easy but is a lot of duplicated code if not done properly. I can do that if required. 

It is working fine on my phone. I would appreciate if someone can try this onto their device. 

Having said that, I am happy about the other ongoing work for conversion to objects and hence didn't spend time to do 1 and 2 as this should be replaced by that PR. So depending upon the progress of the other PR we can chose to ignore this for now also.

My aim was to have a dirty version.

If we decide we will go with this for now, then as this replaces does the following : 
 Android -> Android Multi Reddit
and this will be saved as a User subscription we might have to later on push an update which will remove all these from the user's sidebar when we have the object PR ready. 

Let me know how you guys feel about this.

ccrama:
Have merged. Great work, thank you!

PR #1376, opened 2016-04-11
Change "pts" to color of vote on comments

Nxt3:
On a comment, the vote count is like: "22pts". Currently, if I upvote a comment, the "22" will become orange--but the "pts" won't. This makes it so the entire score span's color is reflected by the vote.


Nxt3:
I also added a space between the score and "pts" text. It looks cleaner this way.

ccrama:
Thanks!

PR #1385, opened 2016-04-12
Improved EditorActions and tweaked layout of editor_items

Nxt3:
EditorActions will now either wrap the text inside of the symbols (i.e. bold or italics) OR insert the symbols and put the text cursor in between the symbols; also, I made it so when you go to enter a list or quote, a blank space is inserted after the list or quote symbol (i.e. "> " vs ">") so the user has one less key to press.

I added a margin to the start and end of the editor items, as it tended to be too close to the dialog edge on a larger screen (specifically, if you had the comment FAB enabled, the popup had its editor actions too close to the left edge).
[Without fix](http://i.imgur.com/AZZIlaO.png)
[With fix](http://i.imgur.com/HLhEHQM.png)

I already tested this on my Pixel C and my Nexus 6P and there doesn't seem to be any weirdness with the editor items when replying via a comment--so you're golden. :+1:

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Views/DoEditorActions.java
+++ b/app/src/main/java/me/ccrama/redditslide/Views/DoEditorActions.java
@@ -52,34 +52,36 @@ import me.ccrama.redditslide.util.SubmissionParser;
  * Created by carlo_000 on 10/18/2015.
  */
 public class DoEditorActions {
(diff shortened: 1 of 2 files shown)


ccrama:
Good find! Didn't have a big enough device to test on but looks better :)

PR #1390, opened 2016-04-12
Improved layout of FAB's for tablets

Nxt3:
FAB on tablets was too far left. This will follow the appropriate keylines given by Google for tablets.


ccrama:
Thanks!

PR #1391, opened 2016-04-12
Fixed synccit posts showing as unread upon a cold boot until refresh()

Nxt3:
Now synccit posts that were sync'd will be added to your seen posts on that device


ccrama:
Sweet, thanks

PR #1394, opened 2016-04-13
Fixed negative scores when downvoting & crash fix

### LINKED ISSUE #1392: If you downvote a post with 0 points, it displays -1. It should still display 0.
Slide version: 5.1.5
Android version: 6.0.1

I tested it out multiple times, I think this is a bug.

Nxt3:
Apparently negative scores aren't a thing for submissions. If a post has a score of "0", keep it at "0" after a downvote. (Fixes #1392)

Also, fixed a crash that happened when sorting by "Controversial past month".


ccrama:
Awesome, thanks

Issue #1395, opened 2016-04-13
Setting for making comment menu "right-handed"


Nxt3:
This would put the upvote/downvote closer to the right side.

Current order:
Upvote, Downvote, Reply, Menu

Proposed order:
Menu, Reply, Upvote, Downvote

PR #1398, opened 2016-04-13
Setting for right-handed comment menu

Nxt3:
Mirrors the comment menu--putting the vote buttons towards the right side of the device.

#1395


ccrama:
Sounds good! Also looking into making a full left-handed mode, but that's more far off :P

Nxt3:
Shouldn't be too hard. Wish you could just "flip" everything programmatically. Essentially forcing LTR layout. I would be surprised if there wasn't a way to do that.

ccrama:
If you set start and end values instead of right and left (ex, paddingStart instead of paddingLeft), you can do it through config changes AFAIK! Would be a lot of converting, though...

Nxt3:
I actually had it on my todo list to accomplish that. I can start on it in the next few days. :)

ccrama:
Awesome, thank you very much!

PR #1407, opened 2016-04-13
Nearly complete coverage of content types (NSFW posts)

Nxt3:
I was able to find posts that weren't tagged correctly--mainly because they were NSFW posts.

I say "nearly" because there may be something that still gets tagged incorrectly. Everything I can find though seems to be correct.

PS: The new way of handling Content types is awesome! Makes adding new ones a breeze.


ccrama:
Sweet, good find! And yeah thank @Alexendoo  for that :)

PR #1410, opened 2016-04-13
Tag external links with the application name that opens them

Alexendoo:
The tagging is memoized per domain as the few ms it takes on the main thread would cause some jank. If there is no app or the URL can't be parsed correctly it uses R.string.type_external

Example:

image


ccrama:
That's actually really awesome! Well done!

Issue #1408, opened 2016-04-13
Reply input field doesn't use wubreddit's accent color and instead uses global accent


The0x539:
![](https://i.imgur.com/9tGT8Gg.png)

/r/grilledcheese has its accent color set to orange for me.

ccrama:
That bar isn't even colored for me?

The0x539:
Is your accent color the whitish gray?

ccrama:
No, it's blue. Only the cursor is colored for me

The0x539:
Even when selected? Odd. Base theme?

ccrama:
Base theme is dark blue

The0x539:
Try the others?

ccrama:
Nope, even on light it's just the cursor. What are your view mode settings and how did you get to that post?

The0x539:
View mode? And I go to the subreddit via navdrawer and then open the post.

ccrama:
Like comment pane, no sub tabs, etc. Might also be a marshmallow thing

The0x539:
Single sub. And yeah, I'm on Marshmallow. I do like it being colored, though.

ccrama:
Huh, this is what I see (and have always seen)

![image](https://cloud.githubusercontent.com/assets/3819844/14506202/36fc1ca4-0182-11e6-80a8-bfde674f4cbd.png)

Nxt3:
It's definitely a Marshmallow thing. I can confirm the save behavior. I'm pretty sure it's because it's using the app's accent color to color it--you don't actually define anything in the code, the EditText field just does that automatically. It should be possible to change it, though: http://stackoverflow.com/questions/26574328/changing-edittext-bottom-line-color-with-appcompat-v7

The0x539:
If it's getting overridden, what are we overriding it to? White, black, or
wubreddit accent?

Nxt3:
We're not overriding it. That's why it's taking the global accent color--because it changes the app's accent color. Marshmallow apparently colors it without doing anything other than using `EditText`

The0x539:
I'm saying when this gets closed.

ccrama:
Just updated to marshmallow a minute ago and I can reproduce now! I'll look at that link later tonight @NXT3

PR #1413, opened 2016-04-14
Fixed bad tinting of editor item icons

### LINKED ISSUE #1396: Editing buttons should have less size and more contrast
...I think? Feels like it, especially on a dark background as far as the contrast goes.

Nxt3:
Fixed #1396 

Made the items have more contrast

### DIFF
--- a/app/src/main/res/layout/editor_items.xml
+++ b/app/src/main/res/layout/editor_items.xml
@@ -18,7 +18,6 @@
             android:cropToPadding="false"
             android:padding="8dp"
             android:scaleType="fitCenter"
-            android:tint="?attr/tint"
             app:srcCompat="@drawable/savecontent" />
 
         <ImageView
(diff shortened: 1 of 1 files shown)


ccrama:
Thank you!

ccrama:
Everything is invisible on the white theme, we need a better solution unfortunately :(

Nxt3:
I tried it with the white theme and it was still visible. Could I have a screenshot?

ccrama:
Reply to a post or edit your reply. The background is white and so are the buttons, unfortunately. I think I'll make a parameter on the DoEditorActions class to force White tinting, might be our best bet

Alexendoo:
Wait so what exactly is the colour set to? If it's white where's that coming from when it was set to the main colour of the sub in #1408

Nxt3:
@Alexendoo `?attr/tint` makes it a color that goes well with the base theme. If the base theme is Light, then it makes the icons "dark"; if the base theme is anything else, it makes the icons "lighter". The issue is that in the above screenshot, he has a dark subreddit theme--but the Light base theme--making the icons dark against the dark subreddit theme. We need a way to change it depending on the subreddit theme.

PR #1415, opened 2016-04-14
Full RTL layout support

Nxt3:
Nothing seems to be borked... and it shouldn't affect 4.2+ :dancer: 

Just make sure we always do this in the future. ;)


ccrama:
Wow this is quite a lot, thank you!

PR #1416, opened 2016-04-14
EditText highlight is now white

### LINKED ISSUE #1408: Reply input field doesn't use wubreddit's accent color and instead uses global accent

Nxt3:
It can be any color, as long as it's white ;)

Fixes #1408

### DIFF
--- a/app/src/main/res/values/style.xml
+++ b/app/src/main/res/values/style.xml
@@ -2,8 +2,6 @@
 <resources xmlns:tools="http://schemas.android.com/tools">
 
     <style name="AppTheme" parent="swipestyle">
-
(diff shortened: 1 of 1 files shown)


ccrama:
Haha, thanks

ccrama:
This has the adverse side effect of changing all switches to white, and also you just had to put that once in the swipeable parent theme instead of every child theme. I'll look for a workaround!

Nxt3:
Darn. :/

Nxt3:
I know how to fix it. Will do it in the morning

PR #1422, opened 2016-04-14
Cleanup in MediaView and GifUtils

Nxt3:
This is lame, but I started working on something else and realized I had no idea what I was doing. Just didn't want my cleanup to go to waste


app/src/main/java/me/ccrama/redditslide/Activities/MediaView.java
```diff
@@ -281,6 +275,7 @@ public void onLoadingComplete(String imageUri, View view, final Bitmap loadedIma
                         Log.v(LogUtil.getTag(), "COULDN'T DOWNLOAD!");
                     }
                 } else {
+                    System.out.println("FUCK");
```
Alexendoo: Probably want to remove that

PR #1430, opened 2016-04-14
Tinted EditText bar so it isn't colorful

Nxt3:
So, after a few hours of playing around with this--the best we can do is use the `?attr/tint` to color the EditText's bar. There isn't a clean easy way to make all of them white without ruining visibility of the EditText field in 'edit_comment`'s layout when using the Light base theme. 

This will make them white for all themes except the Light theme--in which case, it tints them black. It works lol


ccrama:
That's fine, thank you for the work!

PR #1433, opened 2016-04-15
Lots of cleanup to SettingsSubAdapter and Palette

Nxt3:
- Will now only reload/refresh views if the user changed something
    - Prior to this, opening the "Subreddit theme" changer, changing nothing and pressing "ok" would reload the views (MainActivity, Subreddit theme list in Settings, etc.)
    - Prior to this, opening the "Subreddit theme" changer and pressing "cancel" would reload the views (MainActivity, Subreddit theme list in Settings, etc.)
- Made TextViews on subreddit theme colors dialog Buttons
- Lots of code tidying up in the `SettingsSubAdapter` and `Palette` classes :+1:


ccrama:
The reason it reloaded everything is if you have the card mode set to change the background, it had to change that color. Does this commit take that into consideration?

ccrama:
Looks like it all checks out, thank you!

Nxt3:
Welcome!

PR #1439, opened 2016-04-15
[WIP] Use objects for subscriptions

l3d00m:
Using objects for subscriptions gives us multiple opportunities in the feature. It's easier to make subreddits local only (you don't have to remove them every time then), multireddits could be displayed in the mainactivity as well and it makes a lot of complicated code obsolete.  
It's not near to being finished yet & it's completely untested.

#### KNOWN BUGS:
- [ ] It doesn't display the correct subreddits on first start 
- [ ] It does ask if you would like to sub to /r/slide each time on login, also I'm already subbed
- [ ] doesn't display frontpage and all
- [ ] probably a lot of bugs in ReorderSubreddits.java (it's not tested there in any way)

#### TODO:
- [ ] ~~**Port the old subreddit storage method (via Stringbuilder, seperated by commas) in sharedPrefs to the new one**. Currently you have to clear cache to make it working.~~
- [ ] Use FileOutputStream or sqllite instead of SharedPrefs to avoid problems ([Reference](https://stackoverflow.com/questions/7944601/saving-a-hash-map-into-shared-preferences))
- [ ] Add sorting of subscription in the object as well
- [ ] Add a better handling for collections
- [ ] Add multireddits to MainActivity
- [ ] Testing, testing, testing. This changes a lot of underlying code and has therefore a high chance of breaking something
- [ ] Add a syncSubscriptionsOverwrite in MainActivity - NotificationAsync. This will allow an automatic subreddit sync. New subreddits will be then displayed at next startup. It wasn't possible before because it would have overwritten if the user deleted a sub. This is fixed now because that is stored in SubscriptionType.HIDDEN
- [ ] Write tests to confirm that everything is working like it should


l3d00m:
I'll close it as it is too far behind master to merge it (if it would be ready). I won't have any time to work on it in the near future too.

Issue #1369, opened 2016-04-10
Disable (grey out) items in the settings that aren't available


Alexendoo:
Only a couple left I could find:

- [ ] Reddit preferences itself, the category disappears when logged out
- [ ] **Reddit prefs → Hide all NSFW photos:** This is force enabled when "I am over 18..." is disabled, it can only be disabled again when you re-enable over18
- [ ] **General → Notifications:** disappears when logged out

Nxt3:
Are you saying you want the settings for Notifications to be grayed out as opposed to it disappearing when a user is logged out?

l3d00m:
correct @Nxt3

Nxt3:
I can do zat

l3d00m:
zat would be great :+1:

PR #1442, opened 2016-04-15
More settings tweaks

Nxt3:
Addresses #1369 

- Added subheaders in the Reddit Prefs activity


Nxt3:
The settings aren't being grayed out--they are disabled, but there isn't any visual indication that they are disabled. Will fix in a bit.

ccrama:
Awesome, thank you!

PR #1457, opened 2016-04-17
Fixed some settings stuff

Nxt3:
Some of the previews were using the wrong font styles.

Also, if you disabled navbar tinting--the view didn't refresh to account for that change. Now it does.


ccrama:
Great work, thanks!

PR #1458, opened 2016-04-17
Changed style of sorting strings

Nxt3:
Looks like that now. As opposed to "past year", "past 24 hours", etc. I think it looks cleaner.
screenshot_20160417-162119


ccrama:
Yeah that looks much better! Thanks

Alexendoo:
This broke the strings in search and maybe some other places that use just the lone sorting_*

Nxt3:
What do you mean? Could I have an example?

Alexendoo:
image

Nxt3:
I'll look into it. Give me a few.

PR #1469, opened 2016-04-19
Use jitpack for RapidDecoder and ShiftColorPicker

Alexendoo:
Gets rid of a binary file from the repo and removes that last untrusted one, we should be good for fdroid with this?


ccrama:
Yep we should be good to go for F-Droid!

PR #1472, opened 2016-04-19
Fixed Strings whenever frontpage is in the phrase

### LINKED ISSUE #1470: Frontpage 'subreddit' theme dialog shouldn't say /r/frontpage
Visit /r/frontpage in a desktop browser, why don't ya?

Nxt3:
Removed "/r/" when frontpage is the subreddit.

Cleaned up wording and externalized some additional Strings.
- Before: "Content to hide in /r/frontpage"; after: "Content to hide on frontpage"

Fixes #1470

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -652,9 +652,7 @@
     <string name="sub_content_filter">Subreddit content filter</string>
     <string name="submission_share_reddit_url">Share comments</string>
     <string name="filter_content">Filter this content</string>
-    <string name="content_to_hide">Content to hide %1$s %2$s</string>
-    <string name="in">in</string>
-    <string name="on">on</string>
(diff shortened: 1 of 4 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -646,6 +652,9 @@
     <string name="sub_content_filter">Subreddit content filter</string>
     <string name="submission_share_reddit_url">Share comments</string>
     <string name="filter_content">Filter this content</string>
+    <string name="content_to_hide">Content to hide</string>
+    <string name="in">in</string>
+    <string name="on">on</string>
```
Alexendoo: Could you use a getString(id, string) format rather than concatenating? Would be better for other languages

ccrama:
Awesome, thank you!

PR #1492, opened 2016-04-20
Fixed all theme issues regarding editor's EditText fields

Nxt3:
I'd like to start off by saying I tested this on every base theme.

- Changed the "Insert Link" dialog to use Material Dialog builder
- Made positive button color on the dialog dependent on the base theme (this used to stay colored to the accent in some cases--this just makes it a uniform color across all instances of the dialog)
- Have the dialog use a layout.xml as opposed to doing it programmatically

This shouldn't break anything while just making everything more consistent.

PS: I know of the bug where the edit text fields would be missing--and I saw that you fixed it. You shouldn't have needed to remove the style from the `styles.xml`; you only needed to remove the line inside the `amber` style theme--that was my mistake

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Views/DoEditorActions.java
+++ b/app/src/main/java/me/ccrama/redditslide/Views/DoEditorActions.java
@@ -11,6 +11,7 @@ import android.graphics.Color;
 import android.net.Uri;
 import android.os.AsyncTask;
 import android.provider.MediaStore;
+import android.support.annotation.NonNull;
 import android.support.design.widget.Snackbar;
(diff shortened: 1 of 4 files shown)


ccrama:
Why re-add the ReplyEditTextTheme? This breaks on anything below Marshmallow

Nxt3:
Oh I didn't realize it broke on below marshmallow. Every screenshot I saw of the issue looked like it was on Marshmallow lol. My bad, I'll remove it in a bit.

ccrama:
Awesome, thank you!

PR #1497, opened 2016-04-21
Prevent authentication code from leaving user's device

fbis251:
This prevents the authentication code returned by reddit from leaving the user's device via an insecure HTTP request.


ccrama:
Awesome, great find!

PR #1506, opened 2016-04-21
Reduced number of calls to Synccit auth by 1

Nxt3:
Doesn't seem to break anything and I'm still able to register.


ccrama:
Yeah that was most likely a bug. Thanks!

Nxt3:
If you're talking about the bug where you have to press "Save" twice to auth with Synccit--this doesn't fix it. I was actually trying to fix that bug lol. This is just a redundant call.

ccrama:
I was actually not aware of that. I meant that it was redundant and uneccesary was a bug

Nxt3:
Oh. yeah; I'm actually already registered; but if I go to synccit settings and press save--it will wait for a bit then fail. If I press it again, it waits and then auths successfully.

ccrama:
Interesting. Let me know if you find the issue!

PR #1507, opened 2016-04-21
Stopped EditText from autofocusing upon entering certain Settings

Nxt3:
Now when you open SettingsSynccit or SettingsHandling the keyboard doesn't automatically open up and focus on the EditText field.


ccrama:
Ah great! Thanks

PR #1512, opened 2016-04-21
Hide "Mark all as read" button in Inbox when on Sent page

Nxt3:
Clean up and readability, also


ccrama:
What are the changes you did to colorprofile?

Nxt3:
Readability and I made the "share" button a button as opposed to an `ImageView`

ccrama:
Ah great. Thanks!

PR #1513, opened 2016-04-21
Restart InboxPage fragment when Mark All As Read

Nxt3:
Replicated the way we refresh the adapter in MainActivity


ccrama:
Awesome, does it go back to the old page too?

Nxt3:
Yessir. Exactly like MainActivity.

ccrama:
You're the best!

PR #1515, opened 2016-04-21
Made Profile toolbar use a menu

Nxt3:
Before, it was using two `ImageViews`. Now it's using a proper menu. Functionality is retained in regards to hiding the `sort` action on certain tabs.


ccrama:
Awesome, thank you!

PR #1526, opened 2016-04-23
Fixed most of the issues getting title tags

Alexendoo:


ccrama:
Wow that's a lot more elegant and simple. Thank you!

PR #1530, opened 2016-04-23
Share Intent uses EXTRA_SUBJECT for title

fbis251:
Other android applications can understand a share Intent's title passed in via the EXTRA_SUBJECT parameter instead of concatenating them to EXTRA_TEXT


ccrama:
Oh that's much cleaner, thank you very much!

PR #1536, opened 2016-04-24
Altered external domain matching, added upgrade function

Alexendoo:


ccrama:
Looks great, really liking the upgrade class too. Thanks!

PR #1537, opened 2016-04-24
Fixed PTR offset in Single Subreddit view

Nxt3:
If you had tabs enabled, when you were in the SingleSub view--the PTR offset was too much.


ccrama:
Great find, thank you!

PR #1543, opened 2016-04-25
Refactor Paginator limit to Constants class

fbis251:


ccrama:
Awesome, we will definitely start using this file for constants! Thank you

PR #1544, opened 2016-04-25
Cleanup to sidebar

Nxt3:
- Longer flairs now aren't cutoff; adjusted layout of the flair text in the sidebar
- Removed unnecessary code in `doSubSidebar()` in MainActivity
- General code cleanup and readability


ccrama:
Great changes, thank you!

PR #1559, opened 2016-04-25
Prevent app from restarting when debug or cAT package is installed

### LINKED ISSUE #1502: Both Slide and Slide (debug) are restarted when pushing a debug build from AS
I actually commented out the lines in the Manifest for the receiver AND commented out the code inside `onReceive()` in `CheckInstall.java` and this is still happening.

Alexendoo:
Fixes #1502

Not sure what the ccrama.me was for, AFAIK that will never be in a packagename unless you have something like me.ccrama.me, so I'll open a PR for it.

Also - what's the purpose of this class?

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/CheckInstall.java
+++ b/app/src/main/java/me/ccrama/redditslide/CheckInstall.java
@@ -15,7 +15,7 @@ public class CheckInstall extends BroadcastReceiver {
     @Override
     public void onReceive(Context context, Intent intent) {
         String packageName = intent.getDataString();
-        if (((packageName.contains("me.ccrama") || packageName.contains("ccrama.me"))) && !packageName.equals("me.ccrama.redditslide")){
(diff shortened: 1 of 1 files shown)


Nxt3:
`ccrama.me` is used for the YouTube Video plugin

Alexendoo:
I see... well the package name is incorrect in that case but I'll amend it none the less

Nxt3:
I figured. It's too late now though lol

Alexendoo:
Just before that though, do you know exactly what the class' purpose is @Nxt3? If I'm understanding it correctly whenever you would install pro or the youtube plugin it would start the app? That seems like an antifeature

Nxt3:
No, that's correct; it's supposed to restart to account for the plugin.

Example: You install Pro, but the Pro features won't actually start showing up until you close Slide and reopen it. I can personally attest to that happening. This solves that by restarting Slide for you.

I assume something similar happens with the plugin.

ccrama:
Yes @alexendoo, as it has to restart to enable pro features or enable the YouTube integration

PR #1562, opened 2016-04-25
Added Medium font setting for titles

Nxt3:
I think we should remove "Bold" and replace it with "Medium" entirely--but that's up to you.

Medium is a middle-ground between Bold and Regular. I think it looks more "material-y"


ccrama:
Cool, thanks! And I use bold so I want to keep it :P

Nxt3:
I understand lol :P

PR #1564, opened 2016-04-25
Travis CI: don't compile Dev build flavour

Alexendoo:
drops about 5 minutes off the run times for CI

Only difference between Prod and Release is the minSdkVersion, and I can't think of a good reason to run CI on a higher version than we're using for releases


ccrama:
Yeah that shouldn't be an issue, good fix!

PR #1585, opened 2016-04-27
Fixed visuals of Reorder Subreddit view

Nxt3:
- Fixed elevation of toolbars
- Fixed icon size of "second" toolbar
- Changed long press item color to be the accent color of the app (previously, it was actually nearly impossible to see unless you were using the Light base theme)


ccrama:
Really great changes, thank you for the last few commits tonight :)

Nxt3:
No problem sir. 💃

PR #1590, opened 2016-04-27
Make sure we have a valid position to scroll when using PreCachingLayoutManager (fixes #1196)

### LINKED ISSUE #1196: H to V image rotation resets position in Main View
Slide version: 4.7.3
Android version: 5.0

Similar to issue #1067, but I think this is actually a different issue.

In /r/all at any position in the main list (sorted Hot).   In the main list, Rotate phone to H.  View image.  While viewing, rotate phone back to V.  Back out of the picture and find the position in the main view lost.  It resets back to the top of the list.

Note:  This does not fail if the image is viewed from the Vertical position first.   e.g.   In main list, rotated V, click on picture.  Rotate H, then back out.   No problem.

I can recreate this issue with multiple images, but here are 2 chosen at random:
https://www.reddit.com/r/pics/comments/4aeecr/body_scans_of_women_at_250_and_120_pounds/
https://www.reddit.com/r/pics/comments/4af54j/i_am_the_terror_that_flaps_in_the_night/

seizonsenryaku:
LinearLayoutManager.findFirstCompletelyVisibleItemPosition() position is orientation dependent and seems to return the constant NO_POSITION on some cases onConfigurationChanged. 
If that happens, use findFirstVisibleItemPosition() instead.
This fixes #1196.

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Fragments/SubmissionsView.java
+++ b/app/src/main/java/me/ccrama/redditslide/Fragments/SubmissionsView.java
@@ -202,6 +202,9 @@ public class SubmissionsView extends Fragment implements SubmissionDisplay {
                     }
                 } else {
                     i = ((PreCachingLayoutManager) rv.getLayoutManager()).findFirstCompletelyVisibleItemPosition();
+                    if (i == RecyclerView.NO_POSITION) {
+                        i = ((PreCachingLayoutManager) rv.getLayoutManager()).findFirstVisibleItemPosition();
+                    }
                 }
             }
             if (currentOrientation == Configuration.ORIENTATION_LANDSCAPE && SettingValues.tabletUI) {
@@ -228,7 +231,7 @@ public class SubmissionsView extends Fragment implements SubmissionDisplay {
         final Context contextThemeWrapper = new ContextThemeWrapper(getActivity(), new ColorPreferences(inflater.getContext()).getThemeSubreddit(id));
         View v = ((LayoutInflater) contextThemeWrapper.getSystemService(Context.LAYOUT_INFLATER_SERVICE)).inflate(R.layout.fragment_verticalcontent, container, false);
 
-        if(getActivity() instanceof MainActivity){
+        if (getActivity() instanceof MainActivity) {
             v.findViewById(R.id.back).setBackgroundResource(0);
         }
         rv = ((RecyclerView) v.findViewById(R.id.vertical_content));
@@ -396,7 +399,7 @@ public class SubmissionsView extends Fragment implements SubmissionDisplay {
                     down = false;
                 }*///todo For future implementation instead of scrollFlags
 
-                if(recyclerView.getScrollState() == RecyclerView.SCROLL_STATE_DRAGGING) {
+                if (recyclerView.getScrollState() == RecyclerView.SCROLL_STATE_DRAGGING) {
                     diff += dy;
                 } else {
                     diff = 0;


seizonsenryaku:
I see two other possible solutions:
-don't even try to use findFirstCompletelyVisibleItemPosition, always use findFirstVisibleItemPosition.
-don't use PreCachingLayoutManager at all, use CatchStaggeredGridLayoutManager with a single column instead (this may simplify the entire class, requires more work).

ccrama:
This is great, can't believe this was overlooked! I should probably change that in some other places too

Thank you for the debugging and the fix :)

ccrama:
Just saw your last comment. I think switching to the CatchStaggeredGridLayoutManager would be a pretty good solution and simplify a lot of casting and logic throughout submission views

seizonsenryaku:
I thought a little more and I think the real cause of the bug is that when you flip the screen to landscape it is possible that there isn't a single submission view that is entirely visible on screen (the screen contains either two or one partial submission views), so the method returns NO_POSITION.
The findFirstVisibleItemPosition returns views that are only partially visible, so that works.

ccrama:
Yeah that was definitely the issue! Great find. 

Going into the future though, I think your point about switching to a staggered grid layout manager is a valid proposal as well

bantar:
Nice work seizonsenryaku!   Thanks for finding this.   And thanks to ccrama for keeping it open and looking too and deploying a fix.   I'll test it as soon as it's pushed to Alpha and report back.

> there isn't a single submission view that is entirely visible on screen (the screen contains either two or one partial submission views),

When I rotate to H, I've noticed that there is almost never a single image showing in the view.  The only reason that I noticed this is because the target picture is often almost always off the screen by 50-80%, after rotation, making me need to scroll a bit before clicking on it.    

Looking at this closer, when there is an imgur image, the best I can get to show on the screen, when H, is about 90% of one post.   I'd say the slide header takes up 35-40% of the screen, then the picture is shown taking up the other 60-65% and the title text and various buttons are off the bottom of the screen.   If I scroll up, the slide header collapses to about 15-20% and even with a 1 line title, I don't see the full post.   The bottom button row is always clipped.   So, I conclude that my phone can never display a single full image post when H.

I'm downloading an update now.

bantar:
Just tried v5.3.3-alpha.3 and we're still failing.  Not sure if this has the fix or not.

ccrama:
@bantar the fix isn't even merged yet. But yes, the issue was I scrolled back to the first COMPLETELY visible item position, which on your small screen was almost never existent (either the bottom of one and the top of another, never a whole card), and my friend was on list mode so all of the posts were completely visible and it had no issue going between it.

bantar:
OK, just changed to list mode with thumbnail to give the most recently reported failing image another try.    It is working correctly and not resetting the list position.  Tried another, same result.     Then put my settings back to default and tried these 2 again.  Both fail.   

I think we have found the culprit!

ccrama:
@bantar Yep! And the small-ish screen size of the GS5 made it seem like just a GS5 issue, as all my devices have larger screen sizes and I always have cards on cropped mode

bantar:
I too have it set to big picture cropped.  Not sure what this setting does, but it doesn't change this specific problem - on vs off.   The image only fit when I put it to thumbnail only.

ccrama:
@bantar well if I had it uncropped, I would have been able to reproduce it, too

bantar:
I'm afraid to be too giddy, but 5.3.3.alpha5 appears to be working now!   I've tried imgur images, imgur albums and gifs.  Only gave it about 5 minutes of a quick test.   Will be more thorough shortly.   However, from the description, I'm sure that we're on the right track!  Thanks!

bantar:
Final confirmation that we nailed this one!  Thanks again.

PR #1592, opened 2016-04-27
Made selected text white

Nxt3:
When long-pressing for selecting subs--this will change the textcolor to white; upon de-selecting items--the font color will return to `R.attr.font`. This is done for readability since the highlighted color is now the accent color :)


ccrama:
Ah, awesome. Looks better with dark themes now

PR #1593, opened 2016-04-27
Don't use PreCachingLayoutManager in SubmissionDisplay fragments

### LINKED ISSUE #1196: H to V image rotation resets position in Main View

seizonsenryaku:
As discussed in https://github.com/ccrama/Slide/pull/1590.
Lots of changes. I haven't tested this extensively.
This also fixes #1196 for multireddits.

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Fragments/MultiredditView.java
+++ b/app/src/main/java/me/ccrama/redditslide/Fragments/MultiredditView.java
@@ -5,11 +5,11 @@ import android.content.Intent;
 import android.content.res.Configuration;
 import android.graphics.Color;
 import android.os.Bundle;
+import android.support.annotation.NonNull;
 import android.support.design.widget.FloatingActionButton;
 import android.support.design.widget.Snackbar;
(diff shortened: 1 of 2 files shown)


ccrama:
Tested it out and all appears to be working well! Thank you for the added cleanup as well :)

Issue #1595, opened 2016-04-27
Post score does not update instantaneously when upvoting a post, refreshing, and un-upvoting it


Controless:
Slide version: 5.3.3-alpha.3
Android version: 6.0.1

It displays fine if downvoting, refreshing, then un-downvoting. Seems to be only an issue when upvoting.

PR #1596, opened 2016-04-27
Use String.format for number related Strings

Nxt3:
- Will now format numbers to their Locale (For example: 123,456.789 in US locale; 123'456.789 in German locale); this is also _slightly_ more efficient than performing an empty String concat (i.e `"" + intValue`)
- General code cleanup and readability improvements

Only place I didn't do this formatting was in regards to gilding. That will still always show `★ {US locale int}`

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Activities/MainActivity.java
+++ b/app/src/main/java/me/ccrama/redditslide/Activities/MainActivity.java
@@ -2798,6 +2798,7 @@ public class MainActivity extends BaseActivity {
     }
 
     public class AsyncNotificationBadge extends AsyncTask<Void, Void, Void> {
+
         int count;
         int modCount;
 
@@ -2849,6 +2850,8 @@ public class MainActivity extends BaseActivity {
 
         @Override
         protected void onPostExecute(Void aVoid) {
+
+
             if (Authentication.mod) {
                 headerMain.findViewById(R.id.mod).setVisibility(View.VISIBLE);
                 headerMain.findViewById(R.id.mod).setOnClickListener(new View.OnClickListener() {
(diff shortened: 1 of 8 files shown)


Nxt3:
Second commit fixes #1595

ccrama:
Thanks for the fix! As for the second part, it's actually more complicated than that (this can still occur if you aren't the logged in user, but have upvoted something and then come back to the listing). 

Take a look at my solution https://github.com/ccrama/Slide/blob/68098cc98ce53dec504f725600e3d25e7bf0d1e7/app/src/main/java/me/ccrama/redditslide/Adapters/CommentAdapter.java#L702 here

Nxt3:
I didn't realize the issue was occurring with comments also. How is this even an issue when you aren't logged in? If you aren't logged in--you can't vote anyways.

ccrama:
I fixed the comment issue. It doesn't do anything when not logged in (it is NO_VOTE), but if you have upvoted it before going to slide (meaning it has already been shifted by your vote), I offset the vote by +1 or -1. This makes the base amount normal and then adds or subtracts one based on what the vote should be in that method, making it work for both things you have or haven't voted on

Nxt3:
I'm not entirely sure I understand what the issue is. 😕 

However, with your change to the comments--there are bugs regarding the count. I'm having instances where I'll upvote a comment with 1pt--and it turns to 3pts after a refresh. I've been testing it in /r/testslideforreddit, so I know it isn't other people voting on those comments either.

I test the fix I just pushed and it seems to work.

Nxt3:
Scratch that. It doesn't work well enough. Hm.

ccrama:
The issue is the vote count is already affected by your vote before loading content. If it is 1 point and you voted on it, the real vote is 0 and the difference is 1 because of your vote. If it has 2, it's 1, and so on. 

We have to forcibly change the value down depending on the vote returned by the API to get it to work every time (up by one if you downvoted, down by one if you upvoted, or none if you didn't vote), and then apply the user's change to the value

Nxt3:
Try that fix. That seems to be the best. I upvote a post on desktop, then

Nxt3:
Closing this because the PR has effectively changed it's purpose :P Opened up a new PR that works on just this.

PR #1598, opened 2016-04-27
Lots of cleanup with Settings

Nxt3:
- Disable "auto hide nav comments bar" setting if "nav comments" setting is false
- Moved "auto hide nav comments bar" setting to be below "nav comments" setting (layout)
- Changed name of `fastscroll` id to `commentnav` id, in reference to the item in the layout (Did _NOT_ change SharedPref's name) (lots of String refactoring as well)
- Changed name of `navcomments` to `volumenavcomments`
- "Main Theme" and "Post layout" settings now animate sliding-in
- Cleanup and readability as usual


ccrama:
Great stuff, thank you!

PR #1601, opened 2016-04-28
Use String.format for Strings that deal with numbers

Nxt3:
- Will now format numbers to their Locale (For example: 123,456.789 in US locale; 123'456.789 in German locale); this is also _slightly_ more efficient than performing an empty String concat (i.e `"" + intValue`)
- General code cleanup and readability improvements

Only place I didn't do this formatting was in regards to gilding. That will still always show `★ {US locale int}`


Nxt3:
I'm going to merge this since any changes we make while working towards solving #1595, will cause merge conflicts.

PR #1602, opened 2016-04-28
Fixed voting states for Comments and Submissions

### LINKED ISSUE #1595: Post score does not update instantaneously when upvoting a post, refreshing, and un-upvoting it

Nxt3:
Trying to fix #1595 

**Edit 1:** So, the Submission voting states seem to be fixed. The _only_ issue that exists with them is that if you open a submission (in comments) on Slide--then go to another device (not Slide) and upvote/downvote that submission from the other device--and then refresh the comments of that submission--the vote count will be correct along with the correct coloring. However, when you leave the submission's comments, the post won't have any indication it was voted on in the submission view until you refresh the submission view.
Not sure if this is worth fixing given the likeliness of that situation happening--I mean, who has two devices open to the same post simultaneously?

**Edit 2:** It feels hacky [given the `ActionStates` AND checking for values in `up`](https://github.com/Nxt3/Slide/blob/voting-state-fix/app/src/main/java/me/ccrama/redditslide/Adapters/CommentAdapter.java#L728), but this works.

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Adapters/CommentAdapter.java
+++ b/app/src/main/java/me/ccrama/redditslide/Adapters/CommentAdapter.java
@@ -698,16 +698,6 @@ public class CommentAdapter extends RecyclerView.Adapter<RecyclerView.ViewHolder
     public void doScoreText(CommentViewHolder holder, Comment comment, int offset) {
         String spacer = " " + mContext.getString(R.string.submission_properties_seperator_comments) + " ";
         SpannableStringBuilder titleString = new SpannableStringBuilder();
-        int sc = comment.getScore();
-
-        switch (comment.getVote()) {
-            case UPVOTE:
-                sc -= 1;
-                break;
-            case DOWNVOTE:
-                sc += 1;
-                break;
-        }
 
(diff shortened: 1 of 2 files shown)


okungnyo:
So what this bug is is that if you load the comments section of a post in Slide, and then upvote the post using another Reddit app, then refresh the comments section in Slide, Slide will not change the upvote arrow to orange despite the fact that you did upvote it?

Nxt3:
@okungnyo Something like that, yeah.

ccrama:
Does it still need more testing or is it totally complete? If so I can debug a little after class :)

Thank you!

Nxt3:
Give it a whirl. I tested as much as I could and it all seem to work fine--except for the bug I pointed out above.

ccrama:
Looking good! And I think it's just best to ignore that possible issue, it's very niche and would be quite difficult to fix correctly. Besides, you can't do that unless you refresh on your browser either :)

Awesome work!

Nxt3:
I agree. No problem 👯

PR #1605, opened 2016-04-28
Avoid recreating the layout on every configuration change

seizonsenryaku:
Doing this we don't need to keep track of the scroll position manually (StaggeredGridLayoutManager seems to take care of it). This also reduces the GPU/CPU use when doing rotations and avoids creating objects that have to be deleted by the garbage collector at some point.


ccrama:
This is really great work, and looks like it's working well! 

Thank you for the PRs the last few days, really helps a lot

PR #1608, opened 2016-04-28
restartTheme() instead of recreate() when changing base theme

Nxt3:
Now it will scroll back to the tab you were on upon changing base themes.

Works perfectly.

I also delete unused code in SubredditView--it was checking for the base theme menu item in SingleSubredditView--when those menus don't have that item.


ccrama:
Awesome! Thanks

PR #1611, opened 2016-04-28
Hide "Share Slide" menu item if user has Pro installed

Nxt3:


ccrama:
Good idea for pro users!

PR #1619, opened 2016-04-29
Fixed headers not scrolling down

Nxt3:
For some reason, if the `duration` is set to anything more than 0, it won't animate. Setting it to `0` has it working as intended...

Also, why don't you have to call `.start()` on any of those animation calls?


ccrama:
Sorta confused on what this is fixing, this code has been working for a few months now :s

Nxt3:
The further you scroll down when you press the header, it either smoothly scrolls up, or instantly scrolls up. If you scroll way down and the header isn't "expanded" then you press the header, it doesn't "expand".

Nxt3:
https://www.dropbox.com/s/ek6f0rx2k16rqkd/2016_04_29_07_23_20.mp4?dl=0

Nxt3:
This is in regards to the commit you made yesterday. You added this yesterday

ccrama:
Ah, gotcha. Thank you!

PR #1620, opened 2016-04-29
Fixed top margin in Settings - LinearLayout

Nxt3:
While fixing the first issue, it creates a new issue with the content not scrolling behind the toolbar. Not sure what else to do.


ccrama:
Actually all that had to be done was change marginTop to 56 dp instead of 64 dp. Thanks for the PR though, but just fixed on my last commit :(

Nxt3:
And it doesn't look weird to you? It looked fine on my tablet--but on my phone, when scrolling, it looked bad. That fixed it?

ccrama:
Yes, looks fine all around now. About to push it

PR #1622, opened 2016-04-29
Better voting state logic

Nxt3:
Now it accounts for posts and comments you've submitted!


ccrama:
Awesome, looks great!

PR #1623, opened 2016-04-29
Fixed intents firing twice if item is clicked rapidly

### LINKED ISSUE #1621: Multiple posts can be opened if tapped in rapid succession
Tap a post and then tap another one (it is easier if you have two fingers ready). They will both open.

Slide should disallow any other posts to open when a post is being loaded.

Nxt3:
- Also fixes click quickly on a comment before the comment menu has collapse getting funky

Added the class `OnSingleClickListener()`. If multiple clicks happen within the specified amount (`300ms` in this case), they will be ignored.

There is even a `wrap()` function that would allow us to wrap an `OnClickListener()` to become a `OnSingleClickListener()`. I commented it out though, simply because it's easier to change two words in the code for a click listener, than move it out into it's own object. :P 

Fixes #1621

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/util/OnSingleClickListener.java
+++ b/app/src/main/java/me/ccrama/redditslide/util/OnSingleClickListener.java
@@ -1,0 +1,57 @@
+package me.ccrama.redditslide.util;
+
+import android.os.SystemClock;
+import android.util.Config;
+import android.util.Log;
+import android.view.View;
+
+/**
+ * Implementation of {@link View.OnClickListener} that ignores subsequent clicks that happen too quickly after the first one.<br/>
+ * To use this class, implement {@link #onSingleClick(View)} instead of {@link View.OnClickListener#onClick(View)}.
+ */
+public abstract class OnSingleClickListener implements View.OnClickListener {
+
+    private long mLastClickTime;
+    private static final long MIN_DELAY_MS = 300;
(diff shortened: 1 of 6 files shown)


app/src/main/java/me/ccrama/redditslide/util/OnSingleClickListener.java
```diff
@@ -0,0 +1,55 @@
...
+     * Called when a view has been clicked.
+     *
+     * @param v The view that was clicked.
+     */
+    public abstract void onSingleClick(View v);
+
+    @Override
+    public final void onClick(View v) {
+        final long lastClickTime = mLastClickTime;
+        final long now = System.currentTimeMillis();
```
seizonsenryaku: Avoid using System.currentTimeMillis() for this kinds of things, as it is based on the "wall clock" of the phone. Use a monotonic alternative such as SystemClock.uptimeMillis()
See the "Class Overview" in https://developer.android.com/reference/android/os/SystemClock.html for more details.

Nxt3: @seizonsenryaku Thank you! I had no idea.

ccrama:
Has @seizonsenryaku 's suggestion been implemented @Nxt3 ?

Nxt3:
Yep. Just squashed the commits

ccrama:
Awesome! Thank you

PR #1624, opened 2016-04-29
Gray out text if "Never Load LQ images" is checked

Nxt3:
Grays out the "Load lower quality images if possible" text if that option is disabled.


ccrama:
Ah good find!

PR #1627, opened 2016-04-29
Fix fast clicks

seizonsenryaku:
Unfortunately #1623 didn't completely fix the issue #1621.
The different posts have different OnSingleClickListener objects and the mLastClickTime wasn't being shared between them, so the original commit only prevented the same view from having multiples clicks registered.
I was able to open different posts by clicking 717ms after the first click, so 300ms didn't seem enough. I've increased it to 750ms, but as this is shared between the views this may be too high therefore the issue may need a different fix.
I've also changed so mLastClickTime is updated only when a click is registered. That way we can at least make sure the user can have at least a single click registered once every 750ms instead of always having his clicks ignored when clicking too fast.

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/util/OnSingleClickListener.java
+++ b/app/src/main/java/me/ccrama/redditslide/util/OnSingleClickListener.java
@@ -10,10 +10,10 @@ import android.view.View;
  * To use this class, implement {@link #onSingleClick(View)} instead of {@link View.OnClickListener#onClick(View)}.
  */
 public abstract class OnSingleClickListener implements View.OnClickListener {
-
-    private static final long MIN_DELAY_MS = 300;
+    
+    private static final long MIN_DELAY_MS = 750;
     private static final String TAG = OnSingleClickListener.class.getSimpleName();
-    private long mLastClickTime;
+    private static long mLastClickTime;
 
     /**
      * Called when a view has been clicked.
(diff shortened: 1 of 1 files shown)


ccrama:
Interesting find, thank you for the PR! Seems to be working well, and 750ms is the right amount of time

Nxt3:
This will make it frustrating to click on comments. 300 was a nice spot for comments.

ccrama:
Check my reply to the commit

Nxt3:
@seizonsenryaku Why share the timeout between the views? I don't see a scenario where a user tries to  open two posts simultaneously. A double tap is probably going to happen on a single View.

ccrama:
Fat fingers @Nxt3

Nxt3:
¯\_(ツ)_/¯

seizonsenryaku:
@Nxt3 the original issue was about clicking on different views... If this does prove to be frustrating, I do suggest reverting this and looking for a different solution or ignoring the issue for now.

Nxt3:
There isn't another solution. This really is the only way. In all honesty--this has been "broken" since Slide came out--it was only reported recently. I think with this PR set at 300ms--you're fine. You have to make a deliberate attempt to trigger this issue. I don't think it's worth the effort given nobody has brought it up for months. I only even encountered the issue once someone brought it up and I made an attempt to click views in rapid succession.

PR #1628, opened 2016-04-30
Fixed header offset when "load full thread" is clicked

Nxt3:
Fixed tapping on inbox items in quick succession, also


ccrama:
Thought I fixed it, must have slipped through. Thank you!

Nxt3:
No problem :)

PR #1637, opened 2016-04-30
Fixed top margin again

### LINKED ISSUE #1597: Post page top margin should be larger
Slide version: 5.3.3-alpha.4
Android version: 6.0.1

screenshot_20160427-142023

Nxt3:
- Moved screenHeight into the Constants class
- Refactored a bit in regards to calculating the PTR offset
- Uses a percentage now to calculate the top margin in Submissions and other views.

Fixes #1597 
(again)


Alexendoo:
Had to look into doing it using dp rather than a general case estimate, #1638

Issue #1465, opened 2016-04-18
PTR offset is wrong when in landscape


Nxt3:
Edit: Actually, its not. It seems like if you go into landscape, then swipe between tabs, Slide creates the activity oddly. I will post a video soon.

Edit: [Dropbox link to video of issue](https://www.dropbox.com/s/zklz9mw68itm991/2016_04_18_12_03_54.mp4?dl=0)

ccrama:
Seems to be fixed with that commit!

PR #1641, opened 2016-05-01
Cleaned up screenHeight calls for the PTR offset

Nxt3:


ccrama:
Does this fix #1465?

Nxt3:
@ccrama It shouldn't have. That was just a mild refactor.

But you're right lol it fixed it. Maybe making it static fixed it? :P

ccrama:
That's a possibility! Glad it's working though

PR #1652, opened 2016-05-02
Long press on Toolbar for "Go to subreddit"

Nxt3:
Upon long pressing on the toolbar--it will display an `EditText` field that lets the user enter a subreddit name and upon submission, it will take you to the subreddit the user specified.

**TODO**
- [x] ~~Some sort of popup below the toolbar that would hold suggestions like it currently does in the drawer (implement `SideArrayAdapter` or something of the like?)~~
- [x] ~~Make this a setting~~
- [x] ~~Remove code from the drawer in regards to search when this setting is enabled~~

@ccrama It's done! Complete with animations and suggestions. Turned out a _lot_ better than I had in my mind. Give it a try, let me know what you think.


ccrama:
I really like this idea, but don't think it is intuitive enough to totally replace the sidebar search. It would work for users who don't use the sidebar anyway, but the sidebar has become the basis of  navigation for the app and it would be confusing to change this action so late in development imo. 

I can do the suggestions in the toolbar if you make it an AuroCompleteTextView (basically the same code in Submit.java). Also there is no longer a tutorial so that's not necessary any more :)

Thank you!

Nxt3:
I understand. I'll make it a setting then! Won't have much time to work on it until after Finals, but i'll get around to it. As of now--this works at a basic level (no autocomplete obv).

Alexendoo:
Seems overly complex to add a separate UI for something so close to an already existing thing, I guess this is a suggestion from relay users since that's how they do it? Perhaps just have it open the drawer and focus the existing search field (same concept as how chrome tries to train the user that you can type queries and stuff straight into their omnibar)

Nxt3:
@Alexendoo I had no idea Relay does it like this.

I saw someone suggest this on the subreddit. I liked the idea and figured it made more sense to get it out of the way, considering I rarely ever use the "Go to" functionality anyways; I'd rather have a cleaner drawer. :P Give it a whirl though. It's pretty neat and works well. :)

Alexendoo:
I can appreciate that some people would like it, personally I don't like having to reach that far up.

But regardless I thoroughly believe it's best to stick to one pattern and do it well rather than adding multiple ways to do the same thing and options for everything, less you risk becoming confusing and/or a jack of all trades

Nxt3:
@Alexendoo I'm making this is a setting, don't worry.

Alexendoo:
Yeah that comes under the "option for everything", adds more complexity to the code etc, there is a tendency for open source apps to get a rediculous amount of settings

l3d00m:
I would rather implement it fully or not at all. Making everything optional isn't always the best idea.

Nxt3:
Closing because features are bad if every user won't use it.

ccrama:
This is actually very cool and very well done! As always, great work @Nxt3 and happy to release this!

ccrama:
One small issue I've noticed is the sub bar still exists in guest mode, might be related to Authentication.isLoggedIn being false?

Nxt3:
Oh you know, it's probably because I only messed with "drawer logged in". I'm surprised this wouldn't crash it due to the resources never bring found. I'll fix it this evening. :)

ccrama:
Actually it did crash but I fixed the crashing in one of my commits, haha. Couldn't quite figure this one out though, so thank you :)

PR #1656, opened 2016-05-02
Add better username validation

fbis251:


ccrama:
Awesome! That class will be useful :)

PR #1663, opened 2016-05-03
AlbumUtils: Refactor Ion calls to OkHttp

fbis251:


fbis251:
We can actually make OkHttpJson return the correct model class from Imgur's response instead of returning a JsonObject. It should be able to perform the validation on its own too or maybe even just return a JSON String and let the AlbumUtils class try to parse the response as an Imgur model object

fbis251:
We can probably make the HTTP calls in their own methods since we have 3 cases now:

1. The https://imgur.com/ajaxalbums/getimages/ returns valid data
2. Need to call mashape album API
3. Need to call mashape image API

Regardless of which case it is, we can make the HTTP call method return a JSON string for each call, then parseJson() can try to parse the responses as a valid model for each call, so the HTTP code can be simplified and we can make parseJson() do most of the work

Also I was thinking we can probably check to see if a hash is possibly an album has`(hash.length() == IMGUR_ALBUM_HASH_LENGTH`

PR #1681, opened 2016-05-05
"public static" fields should be constant

AymanDF:
This pull request is focused on resolving occurrences of Sonar rule 
squid:S1444 - “"public static" fields should be constant”. 
You can find more information about the issue here: 
https://dev.eclipse.org/sonar/rules/show/squid:S1444
Please let me know if you have any questions.
Ayman Abdelghany.

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/ActionStates.java
+++ b/app/src/main/java/me/ccrama/redditslide/ActionStates.java
@@ -11,12 +11,12 @@ import java.util.ArrayList;
  * Created by carlo_000 on 2/26/2016.
  */
 public class ActionStates {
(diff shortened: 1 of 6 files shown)


ccrama:
Does making the ArrayList final hinder it's ability to be written to at all? It should still use the same memory pointer so probably won't, just would like to confirm before merging, though!

seizonsenryaku:
You can still add/remove objects from the ArrayList, you just can't change the ArrayList that the variable is referring to.

`final ArrayList<Object> a = new ArrayList<>();`
`a.add(new Object()); // this is fine`
`a = new ArrayList<>(); //this gives you an error`

ccrama:
Awesome, just confirming!

PR #1691, opened 2016-05-05
Use the refresh icon rather than replay

Alexendoo:
ie https://design.google.com/icons/#ic_replay vs https://design.google.com/icons/#ic_refresh


ccrama:
Good find :P

Alexendoo:
@alexsmithfanning spotted it

PR #1695, opened 2016-05-05
Long click Multireddit to view public multis

### LINKED ISSUE #1650: Add public multireddits to profile view
It would be really cool if we could view a user's [public multis](https://github.com/thatJavaNerd/JRAW/blob/master/src/main/java/net/dean/jraw/managers/MultiRedditManager.java#L280). I feel this would fit well as a new tab to the right of "Gilded" when viewing a user's profile.

mjwhitta:
Long clicking Multireddit opens a prompt for a username to view that
user's public Multireddits. Can now view public Multireddits when not
logged in. Can also jump to a user's public Multireddits from the
Profile view. If I had to touch code that appeared redundant, I tried
to create shared code. This pull request closes #1650.


ccrama:
Very great work! Looks great and works well. Thank you!

PR #1714, opened 2016-05-08
Setting for Both drawer and toolbar search methods

Nxt3:
- Now the user can have both the drawer and the toolbar search methods active

Also, a bugfix:
- Upon scrolling down in the Main view--if the toolbar search UI is present, it will now close the search UI


ccrama:
Awesome! Good idea

okungnyo:
What does this do exactly?

Nxt3:
Settings > General > Search method

PR #1730, opened 2016-05-10
Tightened up PTR

Nxt3:


ccrama:
It is now behind the toolbar with sub tabs enabled

Nxt3:
Screenshot? Neither of my devices show it

Nxt3:
I fired up an N4 emulator and I see the issue. Give me a few and I'll fix it.

Issue #1658, opened 2016-05-03
Multireddits as collections bugs/suggestions


Controless:
Slide version: 5.3.4
Android version: 6.0.1
- [x] Accent color cannot be changed for multireddits when added as collections (EDIT: in 'Comment pane' or 'No subreddit tabs' mode)
- [x] Change all instances of /r/sub+sub+sub to /m/multi. Examples: http://imgur.com/a/pYDL1
- [x] Remove the /r/ before /m/. Examples: http://imgur.com/a/iwcRK
- [ ] 'Always show big pictures in this subreddit' option doesn't work
- [ ] 'Always show selftext previews in this subreddit' option doesn't work
  Suggestions:
- [x] Disable 'Submit' from the 3 dot menu, since it's not clear what subreddit you're posting to in the multireddit
- [ ] Add a right sidebar displaying the subreddits in the multireddit (exactly how it's implemented on multireddits page) (Duplicate of #1542)
- [x] In a /m/: the accent color for the "tabs" view mode for the selected tab is correct; however, the FAB ignores the accent color of the multi.

Nxt3:
"Accent color cannot be changed for multireddits when added as collections" needs reproduction. I can change the accent colors for all of my multis with no problem.

Controless:
Here's a video of the bug in action: https://streamable.com/vsgw

Nxt3:
It only happens if you have View mode set to "No subreddit tabs" or "Comment pane".

Controless:
Oh yeah, that could be it. I have it set to Comment pane mode.

Controless:
I think @Nxt3 checked "Remove the /r/ before /m/." already but the issue still persists for the second image in this album:Remove the /r/ before /m/.

Nxt3:
I marked it as done. I didn't realize that was an album. Will fix it later.

Tensounder54:
This issue has been resolved and needs to be closed.

PR #1733, opened 2016-05-10
Hide sidebar and submit menu items in various views

Nxt3:
Both of these suggestions are from #1658 

Hide the "Sidebar" and "Submit" overflow menu items if the currently viewed sub is one of the following:
- multi (i.e name contains `"/m/"`) [only while in the Main view, not the MultiReddits view]
- frontpage
- /r/all
- domain (i.e. name contains `"."`)


ccrama:
Awesome! Looks great

PR #1734, opened 2016-05-10
Remove unused imports

Nxt3:
I don't think this should break anything. Submitting as a PR just in case you think it will.


ccrama:
Great cleanup, thanks!

PR #1746, opened 2016-05-11
Adjusted PTR offset since the issue was fixed for the Priv

Nxt3:


l3d00m:
Hey@nxt3,

You could also push commits directly, without making a pr. This would be especially useful for smaller things like this where you're going to merge it directly anyway. The advantage is that there's no merge commit message then.

Nxt3:
I didn't think I was allowed to do this :P

I'll do this in the future--sorry!

Nxt3:
@l3d00m how exactly do I do this?

PR #1749, opened 2016-05-12
Changed external storage flag

alexsmithfanning:
Changed it from "preferExternal" to "auto" so the user has the option to choose where the app should go. At least, that's what I hope this will accomplish.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -14,7 +14,7 @@
         android:fullBackupContent="true"
         android:hardwareAccelerated="true"
         android:icon="@mipmap/ic_launcher"
-        android:installLocation="preferExternal"
+        android:installLocation="auto"
         android:label="@string/app_name"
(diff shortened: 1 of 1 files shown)


l3d00m:
Sounds good to me! 

Here's an explanation for others:

> If you declare "preferExternal", you request that your application be installed on the external storage, but the system does not guarantee that your application will be installed on the external storage. If the external storage is full, the system will install it on the internal storage. The user can also move your application between the two locations.

> If you declare "auto", you indicate that your application may be installed on the external storage, but you don't have a preference of install location. The system will decide where to install your application based on several factors. The user can also move your application between the two locations.

ccrama:
Actually, it was on auto before and was changed to preferExternal so the user could switch it if they wanted

alexsmithfanning:
For some reason I can't switch it. Does it give you the option?

ccrama:
Worth a shot at least, let me know if it shows up after this!

PR #1750, opened 2016-05-12
Search type and time frame are now in the toolbar subtitle

Nxt3:
edited_screenshot_20160512-182016


ccrama:
Looks great! Thanks

Issue #1719, opened 2016-05-09
Blank space appears in comments with code only content.


piyush-rvc:
Slide version:5.3.4
Android version:5.1.1

Initial two lines are blank if the comment contains only code.

ccrama:
Thanks for the report! Duplicate of #1514

piyush-rvc:
@ccrama  I would like to fix it, where should I look into. It seems processing for comment containing code happens some where in addition to commentAdapter, which is causing the issue. Thanks!!

piyush-rvc:
@ccrama  also, why is comment display as firstTextView and CommentOverflow? IMO, using only CommentOverflow would do the job.

ccrama:
@piyush-rvc  thank you! And it's like that because we don't have to re-inflate a new first textview every time the view is displayed, which makes it much faster. The CommentOverflow is only touched when needed, which makes scrolling much faster

PR #1752, opened 2016-05-13
Fix for  Blank space appears in comments with code only content. #1719

piyush-rvc:


piyush-rvc:
this doesn't work for comments. Never mind.

PR #1753, opened 2016-05-13
Fix for :  Blank space appears in comments with code only content. #1719

piyush-rvc:


piyush-rvc:
This is not complete fix. The blank space that appears due to firstTextView has been fixed. There are new lines present in Comment Overflow as well, they need to be investigated separately.

ccrama:
Great find! And those other lines are on purpose (the padding is still there), but that's negligible compared to the height of the firstTextView, which made it look a lot stranger.

piyush-rvc:
@ccrama `appRestart` `sharedPreference` contains `Preference` "back", which doesn't seem to be used anywhere in non redundant faishon. Please check it so that I can clean it up.

ccrama:
Please don't remove either @piyush-rvc , those both have use in the Reddit.java class

piyush-rvc:
"back" is used in Reddit.java, but I can't see how it affects control flow. Also, "firststarting" is not used in Reddit.java @ccrama

ccrama:
They were there for a reason at some point @piyush-rvc , keep them there please. Don't quite remember what for but I'd rather not inadvertently create other issues

piyush-rvc:
@ccrama okay, no problem :)

piyush-rvc:
@ccrama I couldn't find a better place to ask, but why the app needs to be restarted? Can't we fix that, other apps don't do it. When the app restarts, I get "slide is not responding" pop up always.

ccrama:
Because a lot of settings change between users unlike other apps

miladmaz12:
Hey, this is off couple but I received an update to v5.3.9.1 but the version number in app says v5.3.1.9 instead of v5.3.9.1

l3d00m:
The version number was a mistake I think, it should be 4.3.9.1 instead of 4.3.1.9

ccrama:
Whoops

piyush-rvc:
@ccrama what is emote enabled sub reddit? I couldn't find anything useful using google. what are emotes and how are they used in reddit?

l3d00m:
I think the emotes were implemented in #1102 with an explanation.

PR #1767, opened 2016-05-15
Treat bildgur.de as imgur.com

ledoge:
bildgur.de redirects to imgur.com, and it's used on the subreddit /r/kreiswichs. However, these links weren't recognized as imgur links by Slide and opened in the browser instead.


ccrama:
Didn't know that, thanks for the PR!

PR #1779, opened 2016-05-20
Improvements to Sending Inbox messages

Nxt3:
- Doesn't hide the FAB anymore
- Displays a Toast with a message of success or failure
    - if the failure was because an incorrect username was the recipient, the Toast will show a "This user doesn't exist" message
- If sending the message failed for whatever reason, stay on this screen so the user will get another opportunity to possibly correct their error
- When composing a message, `requestFocus()` on the `to` field
- When replying to a message, disable editing in the `to` and `subject` fields
- Layout improvements
- General cleanup to the code


ccrama:
Looking good, thank you for the great work!

Nxt3:
No problem 💃

PR #1781, opened 2016-05-20
Add functionality to remove local subscriptions

joakim5937x:
This will give the user the choise to remove subreddits that isn't
connected to a reddit account.

This will fix the bug where you can't unsubscribe from /r/slideforreddit
when added from first account-login (the alert dialog asking you to subscribe to /r/slideforreddit)


ccrama:
Great work, thank you for the pull request!

Issue #1780, opened 2016-05-20
N-style full width list view


Nxt3:
Someone on the subreddit was saying they'd like to see the full-width style list view. I would add a bit of spacing between the items though.

[See what I'm talking about here](http://i.imgur.com/JbnX82t.jpg).

What are the thoughts on this? I could do it if you decided you liked it, @ccrama.

okungnyo:
That looks insanely good.

Is it a mock up or a real app?

l3d00m:
Looks like a screenshot from the material design specifications.

Nxt3:
@l3d00m is correct. From their specs. Whether or not I will implement this is up to ccrama. :P

Nxt3:
I wasn't going to make it an option--this was going to replace the current style of list view. Thoughts? I could do it as an option if you wanted.

ccrama:
I'd have to see what it looks like first. I personally like the current list view but that could change

Nxt3:
Alright, I'll work on it and post back here once I have it styled. :)

Nxt3:
Here's what I got. The downside is that is the thumbnails larger--we need vector images of the "hide", "nsfw", and "web" drawables. Not sure how to go about that.

I think if we make this the default new list view--we should make the thumbails for the normal card views use the wider styled thumbail--I think it looks better and less "boring". Don't be a square.

![screenshot_20160521-153822](https://cloud.githubusercontent.com/assets/2940826/15450469/3dec65fe-1f6a-11e6-8e9a-074941f1c0f7.png)

lbds137:
Ooh, I like this a lot! Reminds me of why I liked Relay's list view - this is quite similar I think, and that's a good thing.

ccrama:
Looks awesome, done with your last PR @Nxt3

PR #1786, opened 2016-05-22
Full-width list view mode

Nxt3:
- Replaces old list view style
- Thumbnails for card and middle card modes now match the style of the List mode (wider image)
- Higher resolution "NSFW" and "web" images

I gotta say, this looks much better than I had imagined. I like the card views--but I might just switch over to this on my phone.

[Looks like the last two images in this gallery](http://imgur.com/a/bNU2P). I opted for the minimal spacing as I find it better looking and less like the cards view mode.


Nxt3:
Fixes #1780

ccrama:
Looks awesome! Thank you for the PR.

One question though, does it totally replace the list view or is it an alternative to the current list view

Nxt3:
It replaces it. Personally, I don't see the need for an option here; this looks just so much better. If you want it as an option, I can do that.

ccrama:
Just tested it out and looks great! There is a slight padding issue with the left hand thumbnail mode, though.

And personally I don't know if it should be an option, some users like a more compact list view with smaller thumbnails vs a bigger list view, what are your thoughts on this?

Nxt3:
I'll fix the left-hand issue, forgot to test that. 

I dunno. It's still more compact than cards; personally, I never considered using the list view mode because I found it to not look as nice as it could. [As per the post on /r/slideforreddit](https://www.reddit.com/r/slideforreddit/comments/4kfeud/what_is_everyones_thoughts_for_this_new_styling/), nobody seems to ask for it to be an option. If you like it, I say add it without an option. If enough users voice their opinion about it, I will absolutely add it as an option.

ccrama:
Yeah you're right, adding it as an option would be a little confusing as well. If you need help with the left handed thing let me know, otherwise I'll merge when that's done. Thank you!

Nxt3:
Fixed 👍

ccrama:
Sweet, merging now! Thank you :D

PR #1789, opened 2016-05-22
Keep screen awake while viewing media

Nxt3:
When viewing a gif, album, or image--the screen will stay awake. Useful for when reading an image with lots of text.

This doesn't form a wakelock.


ccrama:
Really great idea, thank you for the PR!

PR #1790, opened 2016-05-22
Improvements to gallery mode

Nxt3:
- Comment button and content type appear on same line (at bottom of image)
- semi-transparent bar drawn behind width of image for the buttons
- slightly more spacing between images
- touch feedback for the comments button

screenshot_20160522-152148

### DIFF
--- a/app/src/main/res/layout/gallery_image.xml
+++ b/app/src/main/res/layout/gallery_image.xml
@@ -3,7 +3,7 @@
     android:layout_width="match_parent"
     android:layout_height="wrap_content"
     xmlns:app="http://schemas.android.com/apk/res-auto"
-    android:padding="2dp">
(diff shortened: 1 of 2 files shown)


ccrama:
Looks really nice! I will look into making the black bar invisible on image posts (to see the whole image), but this is great for other content. 

Thank you!

Nxt3:
If you do that, you should put a different background behind the comments button. Maybe set an elevation for it? It becomes impossible to see on white backgrounds otherwise.

Nxt3:
What exactly is that? It's below the image? But we can't do that for Gallery mode since it sits on top of the image.

ccrama:
Not on my phone, it goes on top of the image (smaller display). I did a test version of it in my last commit if you want to check it out

PR #1793, opened 2016-05-23
Add language override on screen rotation 

joakim5937x:
Added the override language function on screen orientation change.

This will fix the bug then the app reverts back to local language on orientation change (in the MainActivity list) although "fore english" is checked


ccrama:
Great solution, thank you very much for the PR!

PR #1794, opened 2016-05-23
Check domain boundaries in host comparisons

Alexendoo:
e.g. `youtube.com` will no longer match `gifyoutube.com`, it's used in both content type tests and postmatch so a user could now add `t.co` without blocking any domains ending with a t on the co TLD. 

It is a stricter match but I can't think of a domain that has multiple non subdomain prefixes where it would cause confusion. It may impact users if they hadn't decided to type out the full domain, but I would imagine that's rare if it happens at all

edit: also this build will fail, because it's already failing, I tested it on non failing version then applied to latest


ccrama:
Much better solution to the current system, thank you very much!

PR #1799, opened 2016-05-24
"Hide read posts" in menu

### LINKED ISSUE #1797: Hide All button when FAB is disabled
If the user has the FAB disabled or set to the "New post" setting--add a menu item for hiding all read posts.

Nxt3:
If the FAB type is set to "Post" or if the FAB is disabled--show a "Hide read posts" menu item.

Fixes #1797


ccrama:
Really good idea, thanks!

Issue #1771, opened 2016-05-16
Can we please have commas for large numbers on Subscriber counts and in link/comment karma?


miladmaz12:
This is not an issue just a minor adjustment.

Nxt3:
It's reported how Reddit reports it.

miladmaz12:
So there isn't a way to make this happen?

Nxt3:
I could see about it. Just so used to it not having commas. It's how all Reddit clients report it to my knowledge.

Controless:
Reddit is Fun uses commas btw.

ccrama:
The issue with this is in Europe, it's a period and not a comma. Probably won't be doing this because of that

miladmaz12:
But I am in the United States and I'm talking about [this](https://imgur.com/SgAuY15)

ccrama:
Yes, but that would not work for Europeans

miladmaz12:
Oh I see. Alright, I'm going to close this, if it's ok with you.

ccrama:
Yeah it's ok, good suggestion though!

Nxt3:
I'm pretty sure we could do this by using `Locale.toString()`; it takes in the default Locale (so if you were European, it would presumably show periods). 

The issue I have with this is I don't care for commas showing up for the score. :|

Controless:
Yeah, it wouldn't look good on the score but it would greatly help for the subscriber number in the sidebar (hard to differentiate 800000 and 8000000 at the first glance).

okungnyo:
Couldn't the region be detected and if it is Europe, then use dots instead of commas?

ccrama:
If there was a way to region detect, yes, but there isn't really

okungnyo:
That's too bad. And having a setting toggle for something so little as this would be way too much, I assume.

Nxt3:
@ccrama You could use `Locale.toString()` like I said :P It does exactly that.

ccrama:
@Nxt3 exactly, but you need to know the current locale first, which is the issue

Nxt3:
@ccrama `Locale.toString(Locale.getDefault(), "%d", score)` will do that. The default Locale is constructed statically at runtime for Slide's process from the device's system settings. If the user is using a European locale on their device--then they'll get the "." instead of ",".

ccrama:
Ah didn't know that! Might look into it again then

ccrama:
[...]

PR #1801, opened 2016-05-26
Use localized comment count

csdodd:
I would like to do this for subscriber count as well but it is more difficult due to it using `String.format` with this string: `%1$d subscribers` and I don't think Android will be happy casting a String to an int. I am uncertain how easy is is to change the String due to the outsourced translation system in place. I also forgot to add `getLocalizedScore` to JRAW. Will do that now.

device-2016-05-26-103145

Partial fix for #1771

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -87,7 +87,7 @@ repositories {
 }
 
 dependencies {
-    compile 'com.github.ccrama:JRAW:a6d8612648'
+    compile 'com.github.ccrama:JRAW:cd2681e9c2'
     compile('com.github.afollestad.material-dialogs:commons:0.8.5.8@aar') {
         transitive = true
     }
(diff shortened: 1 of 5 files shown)


Alexendoo:
Users have the ability to override the locale in Slide to en_US, I believe this wouldn't accommodate that, would be perhaps better to leave the string alone in JRAW and just format it in Slide

ccrama:
That is a good point @Alexendoo, @csdodd we might want to look into that. Your JRAW solution works well for almost every case, but if the user is in Europe and changes it to force to English, it would still show a period instead of a comma

csdodd:
I tried it out, starting the app in a locale that uses space for divider then forcing en_US works as expected: 
no_to_en

However, after disabling en_US the number does not reset until the app is restarted - so there is a bug going the other way.

miladmaz12:
Hey can we do this but with a user's profile link/comment karma as well?

PR #1805, opened 2016-05-26
Huge font size option

Nxt3:
Someone requested a larger size. So, instead of `+2sp` over the Larger size, I made a Yuge option that is `+4sp` over the previous size.


ccrama:
Haha looks good, thanks!

PR #1807, opened 2016-05-27
Added option for small thumbnails and large thumbnails

Nxt3:
Two things I couldn't fix
- When changing the thumbnail size--the card in Layout Settings doesn't update to reflect the changes between the thumbnail size.
- When changing the thumbnail size, upon returning to the `MainActivity`, the view wouldn't refresh--so you'd still have the old setting. I think this happened when switching between the two sizes only. (i.e. Big -> Small == no refresh; Big/Small -> Big picture cropped == refresh). To mitigate this, I just used the `SettingsTheme.changed = true;` workaround to force a refresh. This was a workaround, and not a fix--but I couldn't figure it out. :|

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Activities/EditCardsLayout.java
+++ b/app/src/main/java/me/ccrama/redditslide/Activities/EditCardsLayout.java
@@ -141,8 +141,18 @@ public class EditCardsLayout extends BaseActivityAnim {
             });
         }
         //Pic modes//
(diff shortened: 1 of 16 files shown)


ccrama:
Fixed all the issues by moving the resizing into the CreateCardView.createCard method. Going to do a few more tweaks and merge with master! Thank you :)

Nxt3:
Thank you. :)

PR #1835, opened 2016-06-05
Drawer edge was not being applied and was throwing exception

prashantkhurana:
Was happening because SidebarLayout had no field named mLeftDragger. 

Also, made it from 6 to 3. 6 was too much. It was conflicting with comment pane.  3 is ideal for me. We can lower it down if 3 feels too much. 

`06-05 16:25:27.282 7747-7747/me.ccrama.redditslide.debug V/(MainActivity.java:625): exception while changing side drawer edgejava.lang.NoSuchFieldException: No field mLeftDragger in class Lme/ccrama/redditslide/Views/SidebarLayout; (declaration of 'me.ccrama.redditslide.Views.SidebarLayout' appears in /data/data/me.ccrama.redditslide.debug/files/instant-run/dex/slice-slice_8-classes.dex)
`

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Activities/MainActivity.java
+++ b/app/src/main/java/me/ccrama/redditslide/Activities/MainActivity.java
@@ -151,6 +151,7 @@ public class MainActivity extends BaseActivity {
     static final int INBOX_RESULT = 66;
     static final int RESET_ADAPTER_RESULT = 3;
     static final int SETTINGS_RESULT = 2;
+    public static final int DRAWER_EDGE_FACTOR = 4;
     public static Loader loader;
     public static boolean datasetChanged;
     public boolean singleMode;
@@ -609,7 +610,7 @@ public class MainActivity extends BaseActivity {
         } else {
             drawerLayout = (DrawerLayout) findViewById(R.id.drawer_layout);
             try {
-                Field mDragger = drawerLayout.getClass().getDeclaredField(
+                Field mDragger = drawerLayout.getClass().getSuperclass().getDeclaredField(
                         "mLeftDragger");//mRightDragger for right obviously
                 mDragger.setAccessible(true);
                 ViewDragHelper draggerObj = (ViewDragHelper) mDragger
@@ -620,9 +621,9 @@ public class MainActivity extends BaseActivity {
                 mEdgeSize.setAccessible(true);
                 int edge = mEdgeSize.getInt(draggerObj);
 
-                mEdgeSize.setInt(draggerObj, edge * 6);
+                mEdgeSize.setInt(draggerObj, edge * DRAWER_EDGE_FACTOR);
             } catch (Exception e) {
-
+                LogUtil.e(e + ": Exception thrown while changing navdrawer edge size");
             }
             if (loader != null) {
                 header.setVisibility(View.VISIBLE);


Nxt3:
What does changing the `DrawerEdgeFactor` do?

prashantkhurana:
Yes. You don't need to drag from the very edge to open up the sidebar.

Nxt3:
The sidebar or the navdrawer? Is this affecting the left or right drawer? I assume the right?

prashantkhurana:
Left one. See this thread : https://www.reddit.com/r/slideforreddit/comments/4j575o/app_drawer_is_difficult_to_swipe_open/
And it resolved in 5.3.7 https://www.reddit.com/r/slideforreddit/comments/4jcoaa/v537_released/ 
But I don't think it did anything as we were catching exception. Let me know if I am missing something.

Nxt3:
So this makes it easier to open--you tested this with all of the different view types? (Tabs, Comment Pane, No Tabs)? I'll test this later and if it looks like it works, I'll merge it.

prashantkhurana:
yes exactly. I did test it, but I would say test it once. See if "3" as a factor is good or if we need less than 3. More than 3 conflicts with comment pane as the drawer opens us if you are trying to go back. 3 is optimal for me. But still test it once.

app/src/main/java/me/ccrama/redditslide/Activities/MainActivity.java
```diff
@@ -620,8 +621,10 @@ public void onClick(@NonNull MaterialDialog dialog, @NonNull DialogAction which)
                 mEdgeSize.setAccessible(true);
                 int edge = mEdgeSize.getInt(draggerObj);
 
-                mEdgeSize.setInt(draggerObj, edge * 6);
+                mEdgeSize.setInt(draggerObj, edge * DRAWER_EDGE_FACTOR);
             } catch (Exception e) {
+                LogUtil.v("exception while changing side drawer edge" + e);
```
Nxt3: Change this to `LogUtil.e(e + ": Exception thrown while changing navdrawer edge size)`, and then delete the two white space lines below this (line 627 and 628). After that, we're more than likely good to merge.

prashantkhurana: Done

Nxt3:
I've merged this into another PR. You've made it apparent to me that the navdrawer sensitivity needs to be worked on--but when `commentPager` is enabled, any value other than 0 lets you bring out the navdrawer from comments. I have it set to disable the drawer edge (set to 0) when viewing comments. Unfortunately, I can't figure out how to reset the value once we swipe back from comments.

prashantkhurana:
Thanks. From what I know, that is the intended behaviour. Try with the current release of slide. Use comment pane, open any post, slide from left edge, you will not go back, instead the drawer will be open. This is the intented behaviour and happens as both subs and posts are in the same activity. 

Personally, as you said, I would like when one is in a post, when he slides from the edge, he should go back. But some might argue that doing that now, is taking away functionality. Some people might use the existing design, to go to subs from within a post. 

Hence, I feel that 3 DrawerEdgeFactor is a good enough compromise which allows you to open the drawer , but if you want to go back, just slide from a little further. 

Let me know your thoughts. 

TLDR :  I agree with what you are saying, but my concern is taking away of functionality.

Nxt3:
I think that is a bug. It makes swiping back more difficult. I swore I saw someone report it as a bug.

Also, 3 is too much for non-comment pane view modes. Swiping between subs becomes a little more difficult; 2 is good for that.

prashantkhurana:
yes, the exact number is subjective. 3 works for me. But yes, I can see why 3 might be too much. 
I know it happens because we never leave the MainActivity. There might be some work arounds, but one way to seperate the subreddit's posts view and a post's comments to a different view. 
As to it being a bug, maybe, but I am not sure if everyone would like taking it away now. Sometimes, when I am reading a post, I use it to switch subs. Not overly attached to it, but do find it a little useful.

Nxt3:
Disabling it and then re-enabling it is the fix. Figuring out where to re-enable it is the issue now.

Nxt3:
I actually wasn't able to find anyone who submitted it as a bug; pretty sure you're right in that this is intended. Keeping it at 2 is the way to go though. Works well on my Pixel C and 6P. 👍 Good work and thanks for bringing this up.

prashantkhurana:
Thanks for following up.

PR #1836, opened 2016-06-05
Tweaked expanded_profile header in drawer

Nxt3:
- Removed submit item
- Tweaked order to match the profile screen order (kept "Saved" at top since it would be a commonly accessed item)

**Before**
screen shot 2016-06-05 at 5 02 27 pm

**After**
screen shot 2016-06-05 at 5 00 59 pm

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Activities/MainActivity.java
+++ b/app/src/main/java/me/ccrama/redditslide/Activities/MainActivity.java
@@ -1651,17 +1651,6 @@ public class MainActivity extends BaseActivity {
                     MainActivity.this.startActivity(inte);
                 }
             });
-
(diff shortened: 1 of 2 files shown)


aaronpoweruser:
While you are tweaking, can i suggest an interaction improvement. Pressing the profile item takes you to the profile view but pressing the header expands the header. Both items have the expand arrow, so the actions should be consistent. I understand it was meant to be a shortcut but it should be easier to just go to the pane you want directly.

Nxt3:
Where would the header go exactly? There isn't an activity associated with the header, only the profile. Hence, why these items behave differently.

ccrama:
I like the changes, thank you for the work (and the commits this week)!

PR #1848, opened 2016-06-11
bugfix where some subs don't show up in 'go to sub'

prashantkhurana:
Example : Try searching for music sub in the "go to sub" .


Nxt3:
Not sure I understand what this fixes. Certain subs don't show up because they aren't suggested. We have an internal list of the top (200?) subs that are suggested.

prashantkhurana:
r/Music is one of them (500). Still doesn't show up.

Nxt3:
oh huh. So it's because of capital letters? Makes sense.

prashantkhurana:
Yes, exactly.

PR #1850, opened 2016-06-13
Update Play/F-Droid badges

### LINKED ISSUE #1818: Use new F-Droid badge
See https://f-droid.org/posts/get-it-on-f-droid-badges/.

anowlcalledjosh:
I updated the download badges to the latest versions, and made them the same size.

Previous:

image

New:

image

The large padding around the logos is mandated by [Google's badge policy](https://play.google.com/intl/en_gb/badges/):

> There must be clear space surrounding the badge equal to one-quarter the height of the badge.

The padding above and below can be mitigated somewhat by using line breaks instead of separate paragraphs:

image

so I can add that to this PR if people think it looks better.

Resolves #1818

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ Slide is available on the Google Play Store and F-Droid.
 <a href="https://play.google.com/store/apps/details?id=me.ccrama.redditslide">
     <img alt="Get it on Google Play"
         height="60"
-        src="https://play.google.com/intl/en_us/badges/images/generic/en-play-badge.png" />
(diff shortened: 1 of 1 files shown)


Nxt3:
Personally, I like the third option with the line breaks. Normally I would merge something as trivial as this--but because it's ccrama's repo--I feel better letting him determine what is best. Might not be until this weekend though.

anowlcalledjosh:
I think that looks better too - I've pushed the commit which makes that change.

ccrama:
Looks great, thank you!

PR #1851, opened 2016-06-13
Workaround Layers Crash

### LINKED ISSUE #1843: crash in /r/diy scrolling to 4th post
Slide version: 5.3.9.3
Android version:  5.1.1

happens everytime. here is a list of the posts http://abload.de/image.php?img=rsdiy9baoz.png

seizonsenryaku:
Hopefully fixes #1843 and https://www.reddit.com/r/slideforreddit/comments/4n95hb/slide_crashes_everytime_my_phone_goes_into/

I basically replaced every instance of CardView with a subclass that doesn't use hardware layers. 
Software layers are too slow, so I'm not using any layer at all.
The idea came from here (https://code.google.com/p/android/issues/detail?id=167551).

This may cause some graphic issues if any of those cards are transparent.
If that's case, maybe we should only replace some of the CardView instances.

I'm unsure whether this causes any performance loss. In fact I think this may be faster.

I've not tested this extensively.


seizonsenryaku:
I can easily reproduce the rotation crash by using an large number of columns.

Nxt3:
Does this fix the rotation crash as well? That's going to be hard for us to test--considering ccrama, @l3d00m, @Alexendoo, and I haven't been able to reproduce this on any of our devices. I'll test this for a few days and let you know if I run into any hitches, though. Solid work man :+1:

seizonsenryaku:
Yes, it does fix the rotation crash for me. 
I could easily crash my Galaxy S5 (Snapdragon 801) browsing the frontpage with a large (3+) number of columns, but it works fine for me after this fix.

seizonsenryaku:
I've updated this with a much simpler fix... simply removing the instances of android:layerType="hardware" from the xml layouts is enough to avoid the crashes.

Nxt3:
I'll test this then. Does it affect performance at all for you?

seizonsenryaku:
I didn't notice any difference in performance.

l3d00m:
I'm pretty sure @ccrama added that as a workaround for something, I can't remember for what though.

Nxt3:
@l3d00m I don't think so. I found the commit where he added it: https://github.com/ccrama/Slide/commit/9a49ea5d058f1eea585b2a61eeeea2705348e648

He didn't mention anything about a bug fix... Hm.

seizonsenryaku:
The crashes occur because Android is trying to create huge textures, over the limit for some devices (larger than 4096x4096 for mine). It crashes at the front page if I set to 2 columns right now (uncropped pictures).
I think he probably added those lines to hopefully increase the performance, but I'm a little skeptical whether this really increase it by any noticeable amount.
Google doesn't use layers by default, and the documentation says:  
"Because hardware layers consume video memory, it is highly recommended that you enable them only for the duration of the animation and then disable them after the animation is done."
(https://developer.android.com/guide/topics/graphics/hardware-accel.html)
I'm not sure if the performance improvement (and I'm not sure if there's any in our case here) can offset the overload of creating and keeping those textures in RAM at all times (like the current code does).

Nxt3:
I've been testing it out since yesterday and have found 0 performance loss on my 6P.

c727:
@seizonsenryaku

If you can upload a test build I will test if this fixes #1843

ccrama:
Great work, thank you for the fix! All seems to be working smoothly

PR #1858, opened 2016-06-19
Fix for unsubscribing subs from manage subs page

prashantkhurana:
The old code was using

`public void unsubscribe(String subreddit) throws NetworkException {
        setSubscribed(subreddit, false);
    }`

which by the signature of the method seems that it would do the job. But it wasn't. Upon diving into JRAW I saw that reddit requires the full name of a sub for unsubscribing and full name of a sub is different from the the display name (which is what we were passing). Changed it to pass the sub which fixes it. 
Now we use 

`public void unsubscribe(Subreddit subreddit) throws NetworkException {
        setSubscribed(subreddit, false);
    }`


ccrama:
That is a really good find, thank you very much for the fix!

PR #1872, opened 2016-06-23
Fix the in menu copy text icon to match function of command.

ZeroCool2u:
@Nxt3 This fixes the icon in the drop down menu, such that the copy text icon matches the function of the menu command. See after and before screenshots, respectively, for clarity.
 
newicon
odlicon


Alexendoo:
Where does this menu appear? I can't seem to find it in app 😕

ccrama:
Good find, thank you for the fix!

PR #1876, opened 2016-06-25
only add all the default subs if user has no subscriptions and  list …

prashantkhurana:
…of subs in sidebar is empty. Also prevents them being added again and again


ccrama:
Good idea, thank you!

PR #1877, opened 2016-06-25
sync multi on main page and multireddit tab

prashantkhurana:
If one edits a multi from the multi tab, if they have the multi added in the main activity, that will also get updated simultaneously.


ccrama:
That is great, thank you!

Issue #1862, opened 2016-06-20
Creating and modifying Multis is awkard


kudiaborm:
Slide version:5.3.9.3
Android version: N Preview 4 NPD56N

- [x] The back button (app not the nav bar back button) does nothing in this menu either.
- [x] No confirmation from the "save" icon in the CreateMulti-Activity
- [x] I load a multi, then select the "edit" icon to add threads and it defaults to the first multi alphabetically. At times attempting to type the name of my preferred multi to edit. Edit: The "create multireddit" from the menu seems to work properly,
- [x] The automatically checked subreddits in this menu when trying to modify subs by selecting "plus" fab button do match any actual multi I have created

l3d00m:
> I load a multi, then select the "edit" icon to add threads and it defaults to the first multi alphabetically. At times attempting to type the name of my preferred multi to edit. 

Sorry, I don't get what you mean by this. Can you explain this futher?

kudiaborm:
No longer occurs after wiping data and cache.
Apologies

ccrama:
First one should be fixed as well

PR #1878, opened 2016-06-25
toast on saving multi

prashantkhurana:
Adding a toast when saving muti. 
mentioned in #1862  "No confirmation from the "save" icon in the CreateMulti-Activity"


ccrama:
Nice touch, thanks

PR #1882, opened 2016-06-29
sync multireddit if user changes multireddit on web

prashantkhurana:
this is further work on https://github.com/ccrama/Slide/pull/1877 . that added sync with the edit multi page. this adds sync with multi's changed via the web.


ccrama:
That is very helpful, thank you!

Issue #1735, opened 2016-05-10
swipe up in gallery to get to comments


boyblunder:
Slide version: 5.3.6.1
Android version: 6.0.1

this is a feature request, to be able to swipe up from an image/GIF/etc to access the comments. the current implementation is to swipe from the left, then click the text to go into the comments. implementing this would save a step.

Relay for Reddit has a "message icon" that you can tap within a gallery to access comments, and I think that having this would be more in-line with Slide's gui-awesomeness of swiping everywhere to get to where you need to be.

The0x539:
I'm not *personally* sure this works well with the current UI paradigm, opened content isn't associated much with where it was opened.

Nxt3:
How would this work if you have vertical albums enabled?

ccrama:
Ah true, would not be possible then, unfortunately

ccrama:
I think that is closeable now, the vertical album issue is pretty significant

okungnyo:
Maybe when Force Touch becomes available.

PR #1891, opened 2016-07-07
Go to comments from images

prashantkhurana:
This allows a user to go to comments from images/gifs/streamable/album. 

This is an alternative to #1735 
Also inspired from https://www.reddit.com/r/RelayForReddit/comments/4pxpgf/what_is_your_favourite_feature_in_relay_that/

This is most upvoted feature there. Nothing wrong with copying/taking inspiration from good features 😄  

Example: 

http://imgur.com/atVQpYa

Screenshots for streamable, album and pic
 : 
screenshot_20160706-225159
screenshot_20160706-225131
screenshot_20160706-225052

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Activities/AlbumPager.java
+++ b/app/src/main/java/me/ccrama/redditslide/Activities/AlbumPager.java
@@ -57,6 +57,7 @@ import java.util.List;
 import java.util.UUID;
 
 import me.ccrama.redditslide.Adapters.ImageGridAdapter;
+import me.ccrama.redditslide.Adapters.SubmissionAdapter;
 import me.ccrama.redditslide.ColorPreferences;
 import me.ccrama.redditslide.Fragments.FolderChooserDialogCreate;
 import me.ccrama.redditslide.ImageLoaderUtils;
@@ -85,6 +86,8 @@ import me.ccrama.redditslide.util.SubmissionParser;
  */
 public class AlbumPager extends FullScreenActivity implements FolderChooserDialogCreate.FolderCallback {
 
+    private static int adapterPosition;
+
     @Override
     public boolean onOptionsItemSelected(MenuItem item) {
         int id = item.getItemId();
@@ -96,7 +99,7 @@ public class AlbumPager extends FullScreenActivity implements FolderChooserDialo
             SettingValues.albumSwipe = false;
             SettingValues.prefs.edit().putBoolean(SettingValues.PREF_ALBUM_SWIPE, false).apply();
             Intent i = new Intent(AlbumPager.this, Album.class);
-            i.putExtra("url", getIntent().getExtras().getString("url", ""));
+            i.putExtras(getIntent());
             startActivity(i);
             finish();
         }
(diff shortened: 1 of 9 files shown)


ccrama:
Could you supply a screenshot of what this looks like or how it works? Thank you!

prashantkhurana:
ohh, sorry. I didn't mean to create the PR yet. Give me 20 min. I'll upload the gifs.

prashantkhurana:
@ccrama updated the main comment. 
Please test before merging. I have tested it all scenarios I could think of, but I could have missed some.

ccrama:
I like it and looks good, one idea though is to make it on the left-hand side of the bar always, so it doesn't jump around with the HQ button. Great work overall though!

prashantkhurana:
Thanks. You mean very left side , right? I'll do that.

ccrama:
Yep, thank you!

prashantkhurana:
How does this look? Should I move it a little to the right? or is this good?
screenshot_20160707-121213

ccrama:
I think that looks good!

prashantkhurana:
Done. Had to change LinearLayout to RelativeLayout to achieve it. Have a look just in case.

ccrama:
Fixed a few small issues, otherwise I like it! Thanks

prashantkhurana:
awesome! thanks.

prashantkhurana:
@ccrama I noticed an issue. Due to comments being on the right, it overlays the size of the gif now. What do you think we should do?

Nxt3:
@prashantkhurana Not sure what you mean. The gif size is displayed to the right of the comments button now.

prashantkhurana:
ohh maybe, I did not update. Let me update.

Nxt3:
I think ccrama fixed that.

prashantkhurana:
ahh, awesome. Yaa, I had my version on my phone. did not checkout master.

Nxt3:
This is missing when you view an Album in vertical mode and then tap an image.

prashantkhurana:
I tried to do it. Turned out to be a little tricky as we are two level deep by then (have clicked the on the album and then on an image. So I left it.

Nxt3:
You should add it to the overflow menu on the vertical album page. :D

prashantkhurana:
yeah, that can be done, but no one would use it I am sure. It is easy to slide than to open a menu at the very top of the screen and click on a button. Lets leave it, I would say and if we get a request to add to the overflow menu, we can do that. I am expecting people to say where is it in the vertical view, but not that 'okay fine add it to the menu'  😄

Nxt3:
All I'm saying is people who use vertical album view don't get the benefit of this. 

Also, if you start in horizontal album view, then go to vertical, then back to horizontal--the comment button is missing.

prashantkhurana:
yes, I totally get you and I tried to add it to the individual images on the vertical album, but I couldn't. So I am all for adding it to individual images in the vertical album. But because I can't do that,  I don't think we should add it to the overflow menu to compensate for that, IMHO. But that's just me. 

> Also, if you start in horizontal album view, then go to vertical, then back to horizontal--the comment button is missing. 

Ahh, this can be fixed I think. I'll do that. thanks.

Nxt3:
I'm not sure I see your logic. You can't have this action in one form of the album views, and not the other. It's largely inconsistent. Regardless of view, they should have the ability to do this. If you don't want to add it to the overflow menu, why not add it in the toolbar as an action in vertical album mode? That makes more sense.

prashantkhurana:
Will add it as an item in the toolbar. That makes sense to me.

prashantkhurana:
Although still I would doubt people would use it. It is at the very top of the screen. Nobody on a phone >=5.2 inch is going to do that. And that too is an inconsistency, why bottom everywhere but top at one place.

Nxt3:
If that were true, why even use toolbars ;)

PR #1893, opened 2016-07-08
Changed web icon to compass for links

Nxt3:
A month ago a user suggested we use a different icon. I rather liked his suggestion, so here's a PR for it.

If you don't like it--no big deal. Just wanted to make sure I could say I held my promise to the user.


ccrama:
Interesting, what does it look like though?

Nxt3:
Go to the "Files Changed" for the PR :P

Nxt3:
screenshot_20160708-001108

Alexendoo:
Still IMO draws too much attention, particularly on orange in your example. Perhaps something like [the link icon](https://design.google.com/icons/#ic_link) without the grey background?

Nxt3:
That was my first choice. This was just a users suggestion. :P

I have an idea

- Instead of the grey background--use a tinted (darker) background of that sub's color. (See an example mockup done in Paint below)

mock

Would take some effort considering we'd be drawing a colored rectangle and then we need to center the icon in it for every post. That creates two more views and at least one more layout to draw, though. Maybe there is a better way I'm not thinking about.

Alexendoo:
It would require grabbing the colour either way I think unless we stick with the grey background, since if it's no background you would need to change between dark/light for the different themes. Now web me says since we're using SVG this would be easy there but I have a suspicion on android it may not be so simple

Nxt3:
**Logic for Light themes**
- _Submission **isn't** colored because sub isn't colored_: Use a darker shade of the light theme (so, off white for the white theme, slightly darker sephia for the sephia theme)
- _Submission **is** colored because the sub is colored_: Background uses a darker color of the subreddit theme's primary

**Logic for Black themes**
- _Submission **isn't** colored because sub isn't colored_: Use a lighter background
- _Submission **is** colored because the sub is colored_: Background uses a darker color of the subreddit theme's primary

**Logic for Dark themes**
- _Submission **isn't** colored because sub isn't colored_: Use a darker shade of the dark/blue theme (so, darker blue for the blue theme, darker grey for the dark theme)
- _Submission **is** colored because the sub is colored_: Background uses a darker color of the subreddit theme's primary

That doesn't seem so difficult. I'm mostly worried about performance (given we've added two new views to every submission)--but maybe I'm over thinking it.

Nxt3:
This is technically done. If we were to implement what @Alexendoo suggested with coloring the background, that needs to be done in a separate PR.

Nxt3:
oh boy.

ccrama:
Yikes, what happened? haha

Nxt3:
I was cleaning out my branches and accidently deleted this branch--then it wouldn't allow me to push and then some www.ohshitgit.com moments happened. Figured the easiest thing to do is close this and start over :P

ccrama:
Hahaha, well glad you submitted the other one. Have merged it!

PR #1896, opened 2016-07-08
allows user to vote using comment toolbar

prashantkhurana:
Hi, 

I added a small feature which allows users to upvote and downvote the submission no matter where they are in the thread. Upvote by flinging on the up arrow in the comment navigation toolbar and downvote by flinging on the down arrow.

Usecase: you are reading comments and want to upvote the submission. This requires atleast 2 clicks and is a little hard on big phones. 

Upvoting and downvoting using toolbar is easy as one you don't have to navigate to any place + toolbar is at the bottom. 

Video : https://www.youtube.com/watch?v=2EVJCKBc3Xk

Let me know what you all think.

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Adapters/CommentAdapter.java
+++ b/app/src/main/java/me/ccrama/redditslide/Adapters/CommentAdapter.java
@@ -142,6 +142,7 @@ public class CommentAdapter extends RecyclerView.Adapter<RecyclerView.ViewHolder
     ArrayList<String> replie;
     private String backedText = "";
     private String currentlyEditingId = "";
+    public SubmissionViewHolder submissionViewHolder;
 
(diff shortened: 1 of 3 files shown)


Alexendoo:
Seems like an action that could easily be triggered whilst attempting to scroll, I'm not sure the benefit outweighs that since it seems like a rare situation, and in that case 2 actions isn't such a problem

prashantkhurana:
I tired it a lot. Didn't press it by chance at all.  The toolbar is at the very bottom. I don't think anyone scrolls at the very bottom. And fast scrolling doesn't register as a fling. All I can say is install the PR and try it out, if it happens we will not check it in.

Nxt3:
How does up voting require two clicks?

prashantkhurana:
If you are deep down in a submission, to vote, you need to go up and lose your place in the comments and then click on upvote OR slide back and then click on upvote and then go back in (where you might loose your position again depending on if you have comments pane on or off).

ccrama:
I think it is a cool addition, will merge! Thank you

PR #1903, opened 2016-07-10
better progress dialog increments

prashantkhurana:
Suppose 5 submissions 
Old : 2% 4% 6% 8% and then done
Now : 20% 40% 60% 80% and done


ccrama:
Ah good find, thank you!

PR #1905, opened 2016-07-11
no more infinite loading in offline mode

prashantkhurana:
the horizontal loading bar for comments was loading forever.


ccrama:
Good find, thanks for the PR!

PR #1906, opened 2016-07-11
GifUtils: Use OkHttp instead of Ion

fbis251:


ccrama:
Really great work, this looks much cleaner and sets a precedent that should be followed throughout the app. Thank you!

PR #1913, opened 2016-07-13
Optimized pngs with OptiPNG

seizonsenryaku:
Reduced pngs total file size by ~621KB without quality loss


Alexendoo:
Did you check the result on the final APK? AAPT does some compression and quantization itself

seizonsenryaku:
As far as I know, AAPT only changes the color palette if that can be done without losing quality. OptiPNG does that and also brute forces through possible PNG compression options to further reduce the file size.

But you are right, the APK file is increased with the optimized PNGs... except when you disable the AAPT PNG optimization.

9540731 bytes (optimized PNGs,  AAPT optimization disabled)
9555651 bytes (optimized PNGs)
9555365 bytes (original PNGs)

So yeah, feel free to close this. You can decrease the APK size by a little bit by disabling the AAPT optimization and merging the optimized PNGs, but the difference is too small to bother (~15KB).

ccrama:
Ah, that's unfortunate. Thank you for contributing to Slide, though!

PR #1914, opened 2016-07-13
properly scale images in BigPicture mode (fixes #1857)

### LINKED ISSUE #1857: wrong size of thumbnails
mentioned in #1843

here is the post:^https://www.reddit.com/r/DIY/comments/4ngk92/magic_mirror_kodi_build/

how it should look and how it looks: https://abload.de/img/nowslide0ka2b.png

seizonsenryaku:
I haven't tested too much but I think this is the expected behavior.

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/SubmissionViews/HeaderImageLinkView.java
+++ b/app/src/main/java/me/ccrama/redditslide/SubmissionViews/HeaderImageLinkView.java
@@ -128,16 +128,11 @@ public class HeaderImageLinkView extends RelativeLayout {
                             if (h > 3200) {
                                 backdrop.setLayoutParams(new RelativeLayout.LayoutParams(LayoutParams.MATCH_PARENT, 3200));
                             } else {
-                                backdrop.setLayoutParams(new RelativeLayout.LayoutParams(LayoutParams.MATCH_PARENT, LayoutParams.WRAP_CONTENT));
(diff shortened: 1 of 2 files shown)


seizonsenryaku:
actually, creates another issue(big vertical padding) if the height is >3200... I'll investigate a little more.

seizonsenryaku:
it seems ok now.

ccrama:
That code was actually quite important, it makes sure the final view won't be too large for android to create a layer for. We need to find another workaround for this issue

seizonsenryaku:
Hm... I though the limitation was set because it might be inconvenient when a image has a huge height. With this code the height should be limited by the screen width anyway (the image aspect ratio is being kept).  This might still be a problem when the height is much bigger than the width.
What about the maxHeight attribute that is set to 3200? Isn't this enough to limit the view size?
Also, what do you mean by layers? Is there any layer involved here? I thought there wasn't any involved after #1851 got merged.

ccrama:
No, Android cannot create views larger than about 4000px, so this code limits image height to 3200px. The MaxHeight attribute is reset when we update the view, so it's actually not even used and that attribute could be removed from the HeaderLinkImageView layout. 

And that's interesting, I don't quite know if this was the same layer issue as the one you fixed, we could do an alpha test and see if anyone gets this issue anymore.

ccrama:
I have been using this for a little bit, we actually have to have the image as centerCrop for the comment view, which does not have full sized image height

PR #1919, opened 2016-07-14
AlbumUtils: Use OkHttp instead of Ion

fbis251:


ccrama:
Again, great work. Thank you!

PR #1922, opened 2016-07-15
MediaFragment: Use OkHTTP instead of Ion

fbis251:


ccrama:
Again, amazing work. Thank you!

PR #1932, opened 2016-07-17
maybe this will fix the slowness of the app people are reporting. I n…

prashantkhurana:
…oticed this issue before with volatile, so I removed all of them, but this was left


ccrama:
Just noticed as well, why is it a private static variable? Is it possible to just make it private? And good find, thanks

prashantkhurana:
It's static as it is used in static context. Wish I could make it non static though. Will look into it later. Though I don't suspect this to be the cause of the issue, its just a submission. 

Also, this fix, this did not solve the issue. I am still having the app behave slow.

ccrama:
Ah ok, no worries, that is understandable

PR #1935, opened 2016-07-18
Add a method to open URLs externally

### LINKED ISSUE #810: "open in browser" not working
It opens the current post in a new view. You can actually keep hitting "open in a browser" over and over again so you have to hit back multiple times to get out of the post view.

screenshot_20160115-153745

Alexendoo:
Fixes #810


ccrama:
Great work, thank you!

PR #1938, opened 2016-07-19
add missing break statement

seizonsenryaku:
avoids opening the share window when trying to open externally


ccrama:
Good find, thanks!

PR #1942, opened 2016-07-19
Fix a crash when visiting an external sub as guest

Alexendoo:
subreddit.isUserSubscriber() was returning null


ccrama:
Good find, thanks!

PR #1948, opened 2016-07-20
"Share Image" action now stores images to app cache

Nxt3:
I found it annoying that when I selected "Share Image", it downloaded the image without any indication; if I wanted the image stored for later use, I'd download the image; if I'm sharing an image, I expect it to be temporarily stored on the device.

So, save the image to the cache (_/data/user/0/me.ccrama.redditslide/cache/shared_image_) instead!

- Creates a uniquely named file suffixed with `.png`
- If the `shared_image` dir doesn't exist, create it
- If the `shared_image` dir **does** exist, delete all images in the directory (this avoids unnecessarily bloating the cache size, as only the most recently shared image will exist in the cache at any given time)
- Display a Toast message if there was an error

PS: sorry for all the white space stuff. I ran IntelliJ's Code Formatter. The changes start on line 998 in `MediaView` and also in `AndroidManifest.xml` on line 22.

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Activities/MediaView.java
+++ b/app/src/main/java/me/ccrama/redditslide/Activities/MediaView.java
@@ -20,6 +20,7 @@ import android.os.AsyncTask;
 import android.os.Bundle;
 import android.os.Environment;
 import android.os.Handler;
+import android.support.v4.content.FileProvider;
 import android.support.v4.view.animation.FastOutSlowInInterpolator;
 import android.support.v7.app.NotificationCompat;
 import android.text.Html;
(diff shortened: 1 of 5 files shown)


app/src/main/java/me/ccrama/redditslide/Activities/MediaView.java
```diff
@@ -997,49 +995,67 @@ public void onLoadingComplete(String imageUri, View view, Bitmap loadedImage) {
                 });
     }
 
-
+    /**
```
Nxt3: Changes start here

ccrama:
This is really great and is a great fix to that annoyance. Thank you for the hard work!

Nxt3:
No problem!

PR #1961, opened 2016-07-22
Fix multiple spoilers on the same line not displaying, support more non-title tag spoilers

Alexendoo:
ref: https://www.reddit.com/r/slideforreddit/comments/4tv40m/more_spoiler_support/


ccrama:
Good find, thank you for fixing this and adding the new spoiler types!

PR #1968, opened 2016-07-22
Ignore sbrowser custom tabs

Alexendoo:
See https://www.reddit.com/r/slideforreddit/comments/4u2qmc/issue_with_custom_chrome_tabs/ for the user report


ccrama:
Good find and fix, thanks!

PR #1969, opened 2016-07-23
load more in offline mode

prashantkhurana:
Adds a comment depth option in the offline settings. Also we will try to load 50 comments for offline now.  Marginal increase in caching time as the still only a single http call is made. 

photo_2016-07-23_11-30-33


ccrama:
Awesome work, thank you!

PR #1996, opened 2016-08-07
Improved the layout of the night mode dialog

Nxt3:
- Title color uses font color (was hardcoded white before)
- Tweaks to padding
- Removed "Enabled" text; instead, put the toggle on the same line as the title
- Removed hardcoded strings

screenshot_20160807-105248


ccrama:
I like it, thank you!

PR #1999, opened 2016-08-07
Allows to swipe away from album pager

prashantkhurana:
helps with those 1 image albums as they are easy to swipe now (and in general easy swipe)

Not the world's best animation though, but not something nobody can't live with.


ccrama:
Very nice! Does this still work well with the open to image intent (like if you switch from the vertical view) and the thumbnail view selector?

prashantkhurana:
yes, it does in my testing.

ccrama:
Awesome! Thank you

PR #2000, opened 2016-08-07
Fix Bug while trying to find correct position to insert a sub

prashantkhurana:


ccrama:
Good find, thanks for the fix!

PR #2007, opened 2016-08-08
Fix subreddit sorting

Soberat:
Also fixes subreddits being downcased on next launch


ccrama:
Good find, thank you for this fix!

Issue #1994, opened 2016-08-07
Hardcoded strings


Soberat:
Slide version: https://github.com/ccrama/Slide/commit/899a94f36ba72494eb7bd2f3403f43ec335617d4
Android version: irrelevant

**Mainactivity.java**
- [x] "Gallery mode" in overflow menu (also in few other activites/layouts)
- [x] Content type options in "Cache comments" (I see "Albums" is gone from here for some reason), popup title, save button
- [x] "Flair:" in the sidebar
- [x] "No offline content found" dialog, as well as the button
- [ ] "submission only" and "comments" in offline mode
- [x] "Select flair" for adding a flair to a post
- [x] "Set flair text" if post flair is editable
- [x] "Flair set successfully", "Error setting flair, try again soon" snackbars
- [x] "Subreddit added to /m/", "An error occured, please try again later" snackbars
- [x] "Friends" title
- [x] "Finding moderators" in sidebar
- [x] "/r/sub mods"
- [x] "Add /r/sub to" (multis)
 

**ManageHistory.java**
- [x] "Backup will occur at"
- [x] "No subreddits will back up" (there's a full stop here as well)
- [x] "submission only" and "comments"

**Discover.java**
- [x] Activity title

**ReorderSubreddits.java** 
- [x] All 3 options in the FAB
- [x] "Add & Subscribe" in subreddit adding dialog
- [x] FAQ text 
- [x] Placeholder text in domain adding

**SettingsTheme.java**
- [x] "Night mode" button (string is assigned, but not used"
- [x] Whole "Night mode" dialog, including title, switch, and radio buttons

**SettingsData.java**
- [x] Activity title
- [x] "Enable datasaving settings"

**SettingsFilter.java**
- [x] ~~Placeholder texts~~
- [x] "Flair filters"

**SettingsReddit.java** 
- [x] "Syncing with Reddit" dialog title

Would love to fix them myself but right now Android Studio is giving me a big headache.

Nxt3:
Fixed the night mode strings in #1996

PR #2011, opened 2016-08-09
String fixes

Soberat:
Fixes https://github.com/ccrama/Slide/issues/1994 and everything else I found while I was at it.


ccrama:
That's quite a lot of work, thank you for this!

PR #2014, opened 2016-08-09
Offline mode improvements

### LINKED ISSUE #1990: Allow caching more than 50 parent comments at a time with offline mode
It would be great if this is a user-configurable option in the settings since certain subs such as /r/askreddit can get a lot more than 50 parent comments

Leaving the default at 50 seems like a good compromise, in addition to allowing more or less comments depending on the user's choice

prashantkhurana:
• Uses an executor to do caching. So one can continue to use the app while caching.Aysntask prevented that.
• removed caching progress from UI to a notification. Again, so one can continue to use the app while caching.
• Save individual posts offline
• Option for number of comments to cache. Fix #1990
• Multiple caches in parallel (if cached manually)

screenshot_2016_08_09_15_20_09
screenshot_2016_08_09_15_20_38
screenshot_2016_08_09_15_22_05
screenshot_20160809-153809

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Activities/MainActivity.java
+++ b/app/src/main/java/me/ccrama/redditslide/Activities/MainActivity.java
@@ -113,6 +113,8 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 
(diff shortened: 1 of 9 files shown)


ccrama:
Wow, this is really amazing work. Thank you so much for this! Makes Slide a real offline mode competitor :)

prashantkhurana:
Thanks !

PR #2065, opened 2016-08-21
Improvements to nightmode layout

Nxt3:
- Made the time selections a Spinner
- added an "Ok" button to dismiss the dialog

screenshot_20160821-130650


ccrama:
I like it! I'll add in the new red theme as well :). Thank you for the great work!

PR #2068, opened 2016-08-22
Fixed app crash on devices with SDK level 15-16

freundTech:


ccrama:
Thank you for these fixes! There are a few conflicts but I'll fix them. 

Thank you again!

ccrama:
Merged. Great work!

Issue #2069, opened 2016-08-22
Autoplay videos with sound continue playing in background in shadowbox


m4potofu:
Slide version:5.4.2.2
Android version:6.0.1

Streamable videos for exemple, I don't know if there is other services with sound.
Same issue in comment shadowbox.

For testing purpose : r/soccer seems to use streamable a lot.

Edit: mixtape.moe, vid.me... so  probably all webm should autostop  when swiping to another post

poutros:
I know that this happens with the youtube shadowbox if you use the youtube background playback xposed module (but only the version with notification visible all the time).

ccrama:
Should be fixed now!

PR #2070, opened 2016-08-22
Fixed Issue #2069: Autoplay videos with sound continue playing in background in shadowbox

freundTech:
Should fix issue #2069


ccrama:
Thank you for the fix, working well!

PR #2071, opened 2016-08-22
Fixed notifications showing message from null for messages from subreddits

freundTech:


ccrama:
Looks good, and good workaround! Thank you for the PR

freundTech:
No problem.

Do you know if it's possible to get ModMail from a sub instead of a user?  
If yes this needs to be implemented for ModMail as well.

ccrama:
Hmm, I do not believe so. There is always a user attached to mod actions so this implementation should work under all circumstances.

PR #2072, opened 2016-08-22
Fixed non default subreddits closing themselfes when not logged in

freundTech:
Calling subreddit.isUserSubscriber() when the user is not logged in throws a NullPointerException, which is caught in line 1842 and wrongly interpreted as the activity being killed.


ccrama:
Great find, will push this out ASAP! Thank you

PR #2073, opened 2016-08-23
open mod.reddit.com externally

Alexendoo:
https://www.reddit.com/r/modnews/comments/4z2nic/upcoming_change_updates_to_modredditcom/ takes effect on 2016-08-24, it will be the new fancy mod mail UI for users whose subreddits are entered into the beta

https://youtu.be/7l1vWnCBMYU


ccrama:
Ah great, thank you!

Do you know if they are planning on making an API we could hook into later down the road?

Alexendoo:
Haven't heard anything, not seen how it would be different apart from surface details so far but then I'm not overly familiar with mod mail

Issue #242, opened 2015-10-24
Changing of app icon color


The0x539:
So the app is THOROUGHLY themeable, but the icon is just that same orange. It's fine if that's the default and kind of its identity, but it's possible to make it changeable. For an example of this, look at QKSMS. It's also GPLv3 so one SHOULD be able to take whatever they need directly from there.

l3d00m:
I think it's pretty hard to implement, QKSMS had its problems with it too. It especially doesn't work with custom icon packs.

ccrama:
This is done now thanks to @freundTech !

PR #2075, opened 2016-08-23
Added experimental support for coloring the app icon

freundTech:
Inspired by Issue #242.

Probably needs more testing on different devices.

Optional improvement that could be added later: Detect if phone is rooted and offer to restart homescreen, so user doesn't have to reboot.


ccrama:
Looks great, could we have an opt in in Settings > Main Theme possibly (since it's experimental)? Again, great work, thank you!

Nxt3:
I agree. Add it to Main theme. Put a header down titled "EXPERIMENTAL". If you don't want to do that, I can after this is merged

freundTech:
It already has a switch labelled as experimental in theme settings.
If you look at the commit and scroll all the way down you can see the label and explanation.

ccrama:
That is awesome! Thank you

ccrama:
It seems to remove the launcher icon on install even with the setting disabled, is there a way we could keep the "legacy" icon until the user enables the colored setting?

freundTech:
Oh. I didn't think about that.

As far as I know there is sadly no way to prevent that. It has to do with how the custom Icon system works.
 Instead of a app Icon the app now has Activity-Alias icons. Activity-Aliases (and with them their icons) can be enabled and disabled at runtime, while app icons can only be disabled at buildtime.

The only way to implement this is to remove the app icon completely. Otherwise you will always have the app icon and an activity icon.

Nxt3:
@freundTech This removed the debug icon for the debug builds.

freundTech:
@Nxt3 Found the problem. It's not this PR, it's #2076 (I was sure that debug icon worked when I tested this PR).

That one fixed the icon resolutions to match googles recommendations, but didn't fix the debug icon resolutions.

The resolution mismatch made android use the original icon.
PR incoming.

Issue #1984, opened 2016-08-01
N Keyboard Issues


ccrama:
Android N pre-releases seem to have a lot of ugly keyboard bugs which are NOT fixable by Slide (afaik), and will hopefully be fixed before the next release.

Relevant issues:

#1930 
#1950 
#1951 

Those issues will be closed and merged into this one.

freundTech:
Can be closed. Was fixed by #2079

Issue #1951, opened 2016-07-21
Able to open post / close keyboard while keeping subreddit list open


matty1086:
Hard to explain this one, but can reproduce 100% of the time.

Repro steps:
Open app
Long press on "Frontpage" in the title to bring up the subreddit list
Tap on a post that's in the background
Post loads
Now press down arrow to close the keyboard
The window with the subreddits still appears with no way to close it (See screenshot)

The only way to close this is to swipe back to the main list
Tap in the search list
Tap the X to close the search

or

Turn off the screen / turn on again

Expected result: If i tap on a post, the search / keyboard should close 
If i close the keyboard the search should likely close as well

Alternatively, don't allow tapping on posts while search / subreddit list is open

Screenshot:
![screenshot_20160720-171524](https://cloud.githubusercontent.com/assets/17831198/17007476/e9a2e53c-4e9d-11e6-861d-f3c04b6f5891.png)

Slide version: 5.4 Alpha 2
Android version: N DP4

Nxt3:
This seems to be an Android N issue. I can't reproduce this.

matty1086:
Yep, does appear N only. Just tried on my Nexus 7 still on 6.0.1 and do not have the issue. Thanks!

Nxt3:
There seem to be several keyboard related Nougat issues. Hopefully they are fixed in the final release.

ccrama:
Don't know if there is much we can do on our end

ccrama:
Merging with #1984

Issue #1950, opened 2016-07-21
Different keyboards to search subreddit list


matty1086:
Using the latest Google keyboard (v5.1) when i pull out the sidebar to search for subreddits there, it brings up a keyboard with a number row, which does not allow swipe input. In my experience, the number row pops up when the keyboard thinks the input field is a password field.

If i long press on the subreddit name to bring up the list, instead of number row, it shows the auto complete bar and does allow for swipe input.

Possible to change the input type on the sidebar input to match the input type when long pressing subreddit name so that swipe works?

Check out the following screenshots:

Sidebar - no swipe input allowed:
![screenshot_20160718-170828](https://cloud.githubusercontent.com/assets/17831198/17007366/f38faa04-4e9c-11e6-89ca-4598ca42610f.png)

Long press subreddit - swipe input allowed:
![screenshot_20160718-170846](https://cloud.githubusercontent.com/assets/17831198/17007372/fed9b5e4-4e9c-11e6-8778-8ea54521a605.png)

Slide version: 5.4 Alpha 2
Android version: N DP4

Nxt3:
This is also Nougat specific. :p I'm sorry.

matty1086:
Have you by chance looked at this to see what is causing this (and the issue in the other N specific bug I posted) in the code?
It would be great if these are N specific issues to get them reported while the release is still in preview on the Android bug tracker. I'd love to do it, but without knowing the code behind, I'm afraid I wouldn't be much use in a useful bug report for Google to look at. I think this would be the better route to go rather than hoping that Google fixes something that they may or may not think (or know) is an issue.

I totally understand not wanting to put a bunch of effort towards a preview release (that's what I get when I signed up for previews), but I'm assuming once N is officially released later this year to the masses, that there will be a push to ensure functionality in Slide is compatible with N?

ccrama:
I think it's an issue with the OS itself and not an issue with Slide, there is really nothing we can do to fix these issues. I hope Google fixes it before the final release

ccrama:
Merging with #1984

Issue #1930, opened 2016-07-17
When typing a subreddit name in the hamburger menu it creates a weird scrolling animation


Canaaduh:
Slide version: 5.3.9.5 - alpha3
Android version: N - Beta 4 (June security patch)

This bug only occurs when typing a subreddit in the hamburger menu using a different keyboard other than the Google Keyboard. Currently I'm using Fleksy and the issue is that when you enter a letter in the "go to subreddit" it starts scrolling on its own.

ccrama:
It scrolls to the top, this is intended

Canaaduh:
Here's a video of the issue: 

https://streamable.com/wvwh

ccrama:
Looks like it's an issue with your keyboard more than a Slide issue, especially if it doesn't happen with the stock keyboard

Canaaduh:
Ah okay. This issue occurs with the following keyboards:

Fleksy
Chrooma
Swift key
Swype

Are you able to replicate this bug on your device with a keyboard other than stock?

Alexendoo:
I'm using Fleksy and no issue here, it _could_ be an N issue

ccrama:
Merging with #1984

PR #2079, opened 2016-08-24
[WIP] Support for Android N

freundTech:
Was fixed: 
- [x] App crashing on first start
- [x] App crashing when uploading images
- [x] App crashing when changing theme
- [x] Subreddit shortcuts
- [x] Completely white dialogs
- [x] Typing in drawer search field scrolls up and down
- [x] Can't leave offline mode

I only tested in the android emulator.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ android:
   components:
   - tools
   - platform-tools
-  - build-tools-24.0.1
+  - build-tools-24.0.2
   - android-24
   - extra-android-m2repository
   - extra-google-m2repository
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,8 @@ android:
   components:
   - tools
   - platform-tools
-  - build-tools-23.0.3
-  - android-23
+  - build-tools-24.0.1
+  - android-24
   - extra-android-m2repository
   - extra-google-m2repository
   - extra-google-google_play_services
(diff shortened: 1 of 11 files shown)


Nxt3:
I'd add it myself (but I don't have N on my 6P yet).

Could you add `android:resizeableActivity="true"` to the AndroidManifest? This should stop the toasts saying "This app isn't supported for multiwindow" even though it works.

ccrama:
Thank you very much for all these fixes! Do you think any of these are easy to fix? #1984

If not, that is fine. Thanks again!

freundTech:
#1930 Still a problem
#1950 Can't reproduce.
#1951 Can't even get the long press menu to open (Not even on 6.0). Do I have to enable it in the settings?

Alexendoo:
#1951 is from General → Subreddit search method

freundTech:
Thanks. Also can't reproduce #1951. Popup and keyboard close correctly.

ccrama:
You aren't going to believe this but the fix to the listview focus was this:

            drawerSubList.setFocusable(false);

ccrama:
Fixed with a commit to master, I think the PR is ready to merge now

freundTech:
@ccrama Not sure if this was caused by your commit (I think it was), but now as soon as you open the sidebar it scrolls down to the search instead of having you click on the search box first.

freundTech:
Ok. Now it should be ready for merge.

ccrama:
That is great, thank you very much!

PR #2082, opened 2016-08-24
Improved sorting selector by making time period selection it's own dialog

freundTech:
Instead of one really long list you now have a short list with Hot, New, Rising, Top and Controversial and only if you select Hot or Controversial you get a second dialog/popup to select the time period.


ccrama:
Does this still highlight the currently selected sorting period?

freundTech:
Yes. It still highlights the selected mode, however because it only shows time period after you selected a mode you will only be able to see the current mode and not the current time period. Only after you selected a mode it bring up the time popup, where the selected time is highlighted.

freundTech:
While testing if it really works like I think it works I just found another bug. Will fix as soon as possible.

freundTech:
Now it should work as expected.

Alexendoo:
What's the replacement for? Those won't be present in the strings

freundTech:
They will be ;)
getSortingStrings() puts "» " before the active icon. The function then checks which item starts with "» ", removes the "» " and colors the text. Now they are removed from the item in the displayed list but still present in the list it is compared to.

Alexendoo:
Seems a little strange, is there no cleaner way to do it?

freundTech:
A bit cleaner would be replacing them in the array right after you searched for them.
Or even cleaner have a second function that directly returns the Spannables instead of the strings.
This was just the fastest solution I came up with.

Nxt3:
Make it cleaner then :P You got time.

freundTech:
Much cleaner now. I removed the previous commit.
I replaced the whole 'Add "» " to the front' hack with a cleaner solution. Not just the part I added.

ccrama:
Looks great, thank you!

app/src/main/java/me/ccrama/redditslide/Reddit.java
```diff
@@ -248,18 +253,51 @@ public static String arrayToString(ArrayList<String> array, String separator) {
     public static Integer getSortingId(String subreddit) {
         Sorting sort =
```
prashantkhurana: @freundTech awesome feature ! this is much cleaner than the the long list we had. Small bug, does not work with multireddits added to the sidebar. From what I can see, all we need is to make 'subreddit' all lowercase as all keys in 'sorting' map are lowercase. Same for the time also in the getSortingIdTime() method.

PR #2085, opened 2016-08-25
open out.reddit.com links externally

Alexendoo:
They're redirects to other links for analytics, sometimes somebody will copy/paste one into another window so we should just hand off to the browser


ccrama:
Good idea, thank you!

PR #2094, opened 2016-08-28
Put New Post Notify action in sub sidebar

Nxt3:
- Made the title of the dialog use string formatting
- Put sub in dialog title on newline so it doesn't cutoff weird
- Remove the Be Notified of new Posts from all overflow menus--too crowded

screenshot_20160828-112706

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Activities/MainActivity.java
+++ b/app/src/main/java/me/ccrama/redditslide/Activities/MainActivity.java
@@ -1628,6 +1628,69 @@ public class MainActivity extends BaseActivity {
             }
         }
         {
+            final TextView newPostsNotif = (TextView) findViewById(R.id.notified_new_posts);
+
+            assert newPostsNotif != null;
+            newPostsNotif.setOnClickListener(new View.OnClickListener() {
+                @Override
+                public void onClick(View v) {
+                    final String sub =
+                            ((SubmissionsView) adapter.getCurrentFragment()).posts.subreddit;
+
+                    if (!sub.equalsIgnoreCase("all") && !sub.equalsIgnoreCase("frontpage") &&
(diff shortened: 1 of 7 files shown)


ccrama:
Looks great! Does it show the user if they've already followed that subrddit for notifications?

Nxt3:
It just does everything it did before when it was in the overflow menu;
just moved the code.

Alexendoo:
Text looks a little long could wrap on smaller phones, go for a simple verb like "notify" perhaps?

Nxt3:
"Subreddit moderators" would have the same issue. "Notify" is too ambiguous. I agree that it's too long though. Just not sure how to shorten it and retain the original meaning.

freundTech:
Just tested it in an emulator. Doesn't wrap with 320x480, mdpi. The drawer is much wider though. (about 90% screen width)

Alexendoo:
"subreddit" could be just dropped from the others as it's implied through context

I think notify would be fine coupled with a checkbox to see if it's on or off, especially since it has an interstitial giving more details

Nxt3:
I'll drop the "Subreddit" text from those actions in another commit. I'll put a checkbox.

Nxt3:
How do I find if the subreddit has this enabled? Can't seem to find a way.

Nxt3:
Oh damn. Did not mean to do that.

l3d00m:
Huh, I does ask you if you sure, doesn't it?

Nxt3:
I pushed to the wrong branch and pushed my commit directly. I'll open a new PR. Already reverted those changes.

Nxt3:
Actually I'm not sure how to handle this. What should I do? My changes on that branch don't count towards changes anymore. :|

l3d00m:
@Alexendoo has the greatest knowledge about git, join the IRC :)

PR #2096, opened 2016-08-29
Fixed system font not working

freundTech:
Telling the app to use the system font should now actually work.

https://www.reddit.com/r/slideforreddit/comments/502j9k/slide_is_not_letting_me_use_my_system_font/


ccrama:
Good find, thank you!

Issue #2095, opened 2016-08-28
Slide doesn't open external links with default app


hdmi:
Slide version: v5.4.2.3
Android version: Android 7.0 (stock)
Device: Nexus 6P

Every time I click a twitter link with slide It directly opened with the inbuilt web browser, not the default twitter app (Flamingo in my case). i clicked the option, always open this link externally but it seems to do nothing.

Here is a video of ths problem https://youtu.be/EEa-DYSkn2I

Nxt3:
I've had this happen from time to time; not sure if it's Slide's fault though.

freundTech:
Can you go to settings->Link handling and check if twitter.com was added to the "Force external browser" list, so we know if it's a problem with the "Always open external" button or the feature itself?

hdmi:
@freundTech Yes, this step has solved my problem.

Here is a screenshot of the Link handling list before:
http://i.imgur.com/aIwMXv3.png

Then after adding twitter.com to the list, the links started opening in the default twitter client
http://i.imgur.com/r4KHb1L.png

So the problem that I see here is that Slide for reddit is not handling the subdomains as it should, imo "mobile.twitter.com" should inherit the "twitter.com" domain too.
By the way, Slide also spams to the link handling list the URLs that are already there.

Thanks for the support :)

freundTech:
Oh. I think the problem is that the button adds the currently opened website (mobile.twitter.com) to the list, while the actual link goes to twitter.com, which isn't on the list.

Twitter.com then redirects you to mobile.twitter.com, but at that point the internal browser is already opened, so slide doesn't check again if it should open the link externally.

ccrama:
I think all we need to do to solve that is only store the "twitter.com" part instead of subdomains, would there be a situation where subdomains are actually used?

freundTech:
Yes, with websites like wordpress (username.wordpress.com), github pages (username.github.io) and similar services that give you free webhosting/subdomains.

ccrama:
Merged your PR which should fix this issue, thank you!

PR #2098, opened 2016-08-29
[Opinions needed] Fixed Always open externally not working for pages that redirect the user

freundTech:
Please read issue #2095 first.

This PR would fix that issue by changing the "Always open this host externally" button to add the link the user clicked on to the list instead of the page the user is currently on.

In my opinion that would be the best solution.

Another solution would be to register a listener for when a new page is loaded (redirect, link click in webbrowser, ...) and check the new url against the open external list every time.
The problem with that would be that it would first open the browser before opening the external app instead of opening the external app directly.

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Activities/MakeExternal.java
+++ b/app/src/main/java/me/ccrama/redditslide/Activities/MakeExternal.java
@@ -1,9 +1,6 @@
 package me.ccrama.redditslide.Activities;
 
 import android.app.Activity;
-import android.content.BroadcastReceiver;
-import android.content.Context;
-import android.content.Intent;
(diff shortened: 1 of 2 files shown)


Alexendoo:
Definitely a compromise either way, I think that taking the original link is what I would expect personally to happen more times than the current one, LGTM 👍

Checking the link on every navigation would be a different issue I believe, I'm not sure that one is massively needed but for instance if you're reading an article with twitter links in it, it would be nice if they followed your preferences. Depends how complex it is to implement

ccrama:
Just saw this PR. I think that would solve this issue, I can't think of a reason the user would want the current page to do this

Issue #1832, opened 2016-06-04
Use image, not emoji, for gold indicator


The0x539:
Looks weird on my aunt's S6 the way it is, it's gray and 3D.

Nxt3:
any suggestions? I don't think this will be easy given how we make the gilded status a SpannableText. Doubt we'd be able to use an image.

Nxt3:
Best we'd be able to do (for the easiest fix) is this: http://www.fileformat.info/info/unicode/char/2606/index.htm

But that looks ugly.

Alexendoo:
If the font being used has the full star looking odd it would likely be the same for the empty one, as they're just one code point apart. Roboto doesn't provide a glyph for either unfortunately so it falls back to a local one. The normal one you'd expect to see is provided by noto sans symbols, might be possible to include a copy of the font.

Could also just blame Samsung for letting people set weird fonts as default

Nxt3:
I always blame Samsung.

Nxt3:
I think we should include a copy of the Noto Emoji font file. There is a black and white version available that only weighs in at 248KB.

[Link to fonts](https://www.google.com/get/noto/#emoji-qaae)

l3d00m:
I think 0.25mb is quite a lot for such a small issue, I'm against this.

Alexendoo:
The individual glyph(s) needed could be easily extracted and repackaged, however I think the bigger challenge is setting the fallback font which I couldn't find a good way of doing. That or we would need to patch in the extra glyphs to the included roboto typefaces

Nxt3:
I agree that the size is too large, but we could definitely extract the single glyph. Honestly though, this sounds like a lot of work for a Samsung issue.

freundTech:
Adding an image to a SpannableText is possible and it's not even that hard:

    int fontsize = a.getDimensionPixelSize(R.styleable.FontStyle_font_cardtitle, -1);  
    float aspectRatio = (float) (1.00 * image.getWidth() / image.getHeight());
    image = Bitmap.createScaledBitmap(image,
        (int) Math.ceil(fontsize * aspectRatio),
        (int) Math.ceil(fontsize), true);
    icon.setSpan(new ImageSpan(mContext, image), 0, icon.length(), 0);

If anyone has a recommendation for what icon to use (Original reddit gold icon?) I could implement it.

Nxt3:
[...]

PR #2104, opened 2016-08-31
Switched gilded icon to use a bitmap instead of unicode emoji

freundTech:
Icon needs discussion:


Nxt3:
Needs to be white since it will show up on a yellow background.

freundTech:
With yellow background do you mean the RoundedBackgroundSpan that was over the star emoji?

I completely removed that span and instead put the imagespan there. This is how it looks now: http://i.imgur.com/JHUd2wT.png  
(I will delete the phone once this PR is merged or closed. I couldn't be leaking content from /r/lounge ;) )

Nxt3:
Dang that looks really nice. but I think you're going to have difficulty with it showing on yellow backgrounds. For example, go color a sub yellow and see how it displays if you have "Color Card Everywhere" enabled. Not sure how you want to handle that.

freundTech:
How about adding a small white border?
Still looks good on grey background: http://i.imgur.com/ZXz00SJ.png
And is visible on gold/yellow background: http://i.imgur.com/QRnl739.png
(Maybe a bit bigger border?)

l3d00m:
I think the approach without border looks a little bit cleaner. You could still see a star on a hold background. I don't think that wouldn't be too big of a issue as the yellow background is probably pretty rare.

Nxt3:
Yeah, without the border looks fine. and you can still see the number of times gilded? (If you need an example, go to /r/all and sort by Top > All Time).

Nxt3:
Actually, try it again with a larger boarder. While you could still see the start--you won't see the outline; it should look consistent across all colors--so having it appear differently on yellow doesn't seem like the way to go imo.

freundTech:
Bigger border grey background (with count): http://i.imgur.com/HjmSUlj.jpg
Bigger border yellow background: http://i.imgur.com/dGEPZFW.png

Nxt3:
Do a slightly bigger border and this looks great. The smaller the border, the worse it will look on lower-res devices.

l3d00m:
I'm still on the borderless side, even a bigger border looks not as good as no border imho.

Nxt3:
I think it looks better without a border--but to say Yellow is "rare" shouldn't be assumed. Keeping it looking visually identical among all colors should be the goal--as that's how it was done before this PR.

freundTech:
While I agree that it's better if it looks the same on all colors that's not at all how it was done before this PR.

Before thos PR it was a white emoji start with a md_orange_500 RoundedBackgroundSpan.

If you have a sub with color md_orange_500 you would also see just the start.

Nxt3:
Oh duh, you're right. In that case, it should be a chance to improve upon that. However, if the general consensus is that it shouldn't have a border--then so be it. Not my choice to make; it's up to @ccrama.

l3d00m:
@freundTech is correct [here](https://goo.gl/photos/qUoEA9ixMYYvJ2pE7) is an example. It's better to make it look differently for some users than to make it (at least in my opionion) bader looking for all users.

Alexendoo:
rel: #1832

ccrama:
I like it, merging now! Thank you for the hard work :)

Nxt3:
@freundTech The gilded icon isn't centered with the comment info span https://imgur.com/tcpdrCT

Issue #2103, opened 2016-08-31
Better handling of /r/random


Nxt3:
This is somewhere Slide needs improvement. The current implementation of navigating to /r/random is fine--until you refresh the posts, and are navigated away from that sub; not to mention, sorting the posts also causes this to occur. 

- [ ] When navigating to /r/random, the toolbar title needs to be the name of the sub you're viewing instead of "random"
- [x] Ability to add a link to /r/random to my sub list. Currently, adding /r/random to the sub list causes a random sub to be added instead
- [x] Don't go to a new sub when sorting posts in the current random sub
- [x] All of the above mentioned should work for /r/randnsfw as well

touzainanboku:
3 seems to be fixed in the latest update (both for random and randnsfw). 1 is partially fixed. It shows random/randnsfw when you initially open it, but if you change the sort mode it shows the subreddit name.

ccrama:
Fixed all except for the toolbar one

Nxt3:
That's good enough then. Being able to refresh for new subs is actually pretty awesome--and the toolbar title should tell you that you're doing random searches. So, all in all--thank you! :D

ccrama:
You're welcome! And do you think it's an issue if you change sorting and the toolbar changes? Might tell the user it's switched to that subreddit instead of still being in "random" mode, or it could be seen as a bug :P

Nxt3:
Oh you know I didn't notice that. I feel like if you're sorting, your okay with browsing that sub as if it were the single sub view for a defined sub. I think it's fine. :)

ccrama:
Cool, not a bug haha!

PR #2107, opened 2016-09-01
Allow adding random subreddits to subscriptions

Alexendoo:
see #2103


ccrama:
Great fix, thank you!

PR #2109, opened 2016-09-01
Replace resource string concatenation with string parameters

Alexendoo:
c01a7690cdd083c8f8cee1adb7551a52d39f60e7 is the relevant commit, the other is a required translation update


ccrama:
Really great work, thank you!

PR #2112, opened 2016-09-01
Speed increase

prashantkhurana:
We have had a fair share of posts with feedback that Slide is slow. A month ago, some changes were made which resolved most of the issues. 

Latest thread is from 2 days ago : https://www.reddit.com/r/slideforreddit/comments/50ck12/slide_is_so_slow_now_a_days/

Though I agree that many a times cache is large, etc is the cause and clearing cache, fixes the issue, but I personally feel some things in slide were slow. 

So digging into it, I fixed a few issues. 

Results : 

-If you are using comment pane and I assume many people use it, then this loading a sub is 2x faster now. In other cases, there should be no effect. 

- Sorting. Sorting posts in a sub, is now 3-4x times faster, sometimes even 5x. 

When I say x times faster, it is that the number of network calls have reduced by x times and assuming network calls is the biggest factor in latency.

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Activities/MainActivity.java
+++ b/app/src/main/java/me/ccrama/redditslide/Activities/MainActivity.java
@@ -206,7 +206,8 @@ public class MainActivity extends BaseActivity {
     boolean                     currentlySubbed;
     int                         back;
     private AsyncGetSubreddit mAsyncGetSubreddit = null;
-    private int headerHeight; //height of the header
+    private int     headerHeight; //height of the header
+    public int     reloadItemNumber = -2;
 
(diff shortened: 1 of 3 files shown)


ccrama:
Really great finds, thank you for investigating and thank you for this PR! Looks like it will solve those two issues well

Nxt3:
Just gotta say--this is incredible. No joke--comments are already loaded before the swiping opening animation is completed. It's insane how much faster this is. Good work @prashantkhurana! :D

prashantkhurana:
Thanks ! . Really nice to hear :)  Didn't know it would help with other modes too.

Issue #2130, opened 2016-09-07
Downloaded album does not keep images order


Geobert:
Slide version: 5.4.2.3
Android version: 6 (vanilla, Nexus 6P)

When downloading an album, the images are neither date sorted (even inverted) or name sorted.

ccrama:
There is really not a way to fix that, Slide downloads multiple at a time and it often loads some images faster than others

Geobert:
as said on reddit, there's a way by adding 3 digits to the hash name ^^

ccrama:
I actually already did that for alpha 2 :P

Forgot to update this

Geobert:
Just tested it, something seems wrong: many pics have the same 3 digits prefix.

Which file is the code of this to see if I can help in the debug ^^

ccrama:
That would probably be inherant with how this works, it adds 001, 002, 003, etc to the beginning of the random UUID I use as the identifier. If you download 3 albums, there will be 3 001 files, but sorted by date and name, it will be in order. Really don't see a way around this

Geobert:
I've tested only with 1 album, that's why I think something's fishy here :) I can help (I was Android dev not so long ago :))

ccrama:
Ah alright, all of the image loading code can be found [here](https://github.com/ccrama/Slide/blob/4cd435fd4eeb0fe0ce47708b31bdad599193d789/app/src/main/java/me/ccrama/redditslide/Notifications/ImageDownloadNotificationService.java#L155), if you would like to take a look, that would be great! Thank you

Geobert:
I found the cause: a service has only one instance, so when receiving multiple intents, the index got override by the last intent. It works sometime when the job has finished then another index is used. I'm looking for a way to fix it.

Geobert:
Yay, it works, pull request created! :D

ccrama:
Should be fixed now!

Geobert:
Cool! I've also seen some usage of the service in for loop without index, is that intended? (Album.java, MediaViewer.java and Tumblr.java)

PR #2149, opened 2016-09-10
Fix images order when downloading albums (#2130)

Geobert:
Fixed!


ccrama:
Great find! Thank you for the fix :)

PR #2159, opened 2016-09-13
Use a stable version of the android gradle plugin

Alexendoo:
The rc version results in issues for those on a stable version of
android studio


ccrama:
Thanks for the info, hope it goes to stable soon

PR #2161, opened 2016-09-15
Changed web icon to link icon

Nxt3:
Given Reddit's recent changes to their iconography--I felt this was a fitting change.


ccrama:
I agree, will merge now

PR #2170, opened 2016-09-17
Fix full text stikethrough formatting

Deadleg:
The upperbound for i was set to one too low, which meant if the ]d]]
marker was at the end of the text the loop would stop at the end of
text, just before the ].

One case where it broke is in the first table of [this post](https://www.reddit.com/r/GlobalOffensive/comments/4kjze9/virtuspro_vs_natus_vincere_starladder_ileague/).


ccrama:
Good find! Thank you for the fix

PR #2171, opened 2016-09-18
Fix table alignment when specified

Deadleg:
The html elements are fairly well formed where they just take the
form <td align="*">. Just need to match for the each case and set
the TextView gravity accordingly.


ccrama:
Thank you! Works and looks great :)

Glad to see you around as well!

PR #2200, opened 2016-09-28
If the user is a known mod, display the Mod drawer item quickly

Nxt3:
Currently it only displays the Moderator drawer item in a `onPostExecute()`. This causes the list to jump around once the task is completed.

With this PR, the Moderation drawer item will be shown instantly--doing the following check to determine if it should be shown at all: `modOf != null && !modOf.isEmpty() && Authentication.mod`. The `onPostExecute()` still changes the visibility to `VISIBLE` in case the user became a mod since the last Authentication; in which case, the list will jump but that will be the first/last time it jumps!


ccrama:
That's a great idea, thank you!

PR #2202, opened 2016-09-28
Made long press duration longer when Peek is disabled

Nxt3:
- Code formatting fixes as well. Just did the following to both `onTouchEvent()`

```
if (SettingValues.peek) {
    handler.postDelayed(longClicked,
                            android.view.ViewConfiguration.getTapTimeout() + 50);
} else {
    handler.postDelayed(longClicked,
                            android.view.ViewConfiguration.getTapTimeout() + 250);
}
```

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/SubmissionViews/HeaderImageLinkView.java
+++ b/app/src/main/java/me/ccrama/redditslide/SubmissionViews/HeaderImageLinkView.java
@@ -54,15 +54,34 @@ import me.ccrama.redditslide.util.NetworkUtil;
  * Created by carlo_000 on 2/7/2016.
  */
 public class HeaderImageLinkView extends RelativeLayout {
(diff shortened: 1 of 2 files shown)


ccrama:
I don't know if I agree with that fully. I like tapping for a bit longer than a tap but not quite a full long press amount

Nxt3:
This really just reverts the old long press behavior when peek is disabled. When peek is enabled--everything you did is the same. When it's disabled, it's just how it used to be. I feel like if you have that particular thing change people are going to complain; mainly because it's too easy to trigger. I agree with you though.

ccrama:
Ah sorry didn't fully understand it. I do agree with you on that, I forgot I turned the long presses into gesture detectors!

Nxt3:
No worries! My bad. Super stoked for this feature. 👍

PR #2208, opened 2016-10-01
Ignore amp. subdomain in reddit links

Alexendoo:
Bug: https://redd.it/55e2sq

I don't believe the subdomain will produce any special links that need to be handled/ignored explicitly

https://www.reddit.com/r/changelog/comments/53q42z/read_reddit_faster_via_google_with_amp/


ccrama:
Interesting bug, thank you for the fix!

PR #2227, opened 2016-10-10
Add gradle plugin for checking dependency versions

Jawnnypoo:
Title says it all!

`./gradlew dependencyUpdates`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,10 +1,13 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
+apply plugin: 'com.github.ben-manes.versions'
 buildscript {
     repositories {
         jcenter()
     }
     dependencies {
         classpath 'com.android.tools.build:gradle:2.1.3'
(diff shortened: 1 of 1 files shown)


Alexendoo:
Android studio checks for updates and highlights those needing it for most of the repos, does this work for new versions of jitpack dependencies also? I believe that's the only one it can't reasonably cover

Jawnnypoo:
Android Studio actually does it for very few libraries (mostly just the support libraries, Google Play Services, and some Google libraries such as GSON, in my experience). Yes, it works for jitpack so long as the dependency has been fetched by at least one person since being tagged in GitHub

ccrama:
That is a really cool plugin, thank you for implementing it!

Jawnnypoo:
Ahh yes, makes sense since the other PRs were accepted. I'll get that fixed later today!

ccrama:
No worries, thank you!

PR #2251, opened 2016-10-23
replace legacy apache httpclient with okhttp

marwanad:
The following PR replaces the legacy http client with OkHttp. 

Also, Limited the access to some variables in the process.

The PR also fixes a bug with the Synccit integration, under `SettingsSynccit`. To reproduce:

1. Login in successfully with a Synccit account.
2. Disconnect Synccit
3. Login with empty fields/invalid accounts.
4. Expected: Receive connection error dialog, Actual: The integration activation popup dialog shows.

This is because the static list `visitedIds` still contains the member "16noez". In that case, I'm just removing it from the list. 

I tried to limit the scope of this PR so as not to introduce breaking changes, ie. keeping the AsyncTasks as is and doing the OkHttp calls synchronously in there, using same protocol as before, timeout values, etc. It would be great to consider an okHttp and RxJava for network calls in the future since it would simplify a lot of the stuff done here.
Targets #2251

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -14,7 +14,6 @@ android {
         // Translations are crowd-sourced
         disable 'MissingTranslation'
     }
-    useLibrary 'org.apache.http.legacy'
     defaultConfig {
         applicationId "me.ccrama.redditslide"
         targetSdkVersion 24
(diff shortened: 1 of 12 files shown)


ccrama:
Thank you very much for this PR! Definitely a better solution, and the decreased dependency is great as well. 

Also, thank you for fixing that bug! 

As for RxJava, I've been looking into switching to it for a few months now, it would just be an immense undertaking as a lot of Slide is built around AsyncTasks. This will definitely be an improvement I will be working on over the next few months, though!

Issue #2193, opened 2016-09-25
Hide notification bar (immersive mode, or fullscreen)


loptow:
I just installed the power tools app to accomplish this, and I was wondering if this would be easy to implement, or something that you could do?

Power tools does an okay job, but having it built into the app would probably be the better user experience.

Flow for reddit (http://i.imgur.com/s2xHGxF.jpg) can be toggled to hide the notification bar. Power tools can actually hide the nav bar down below too, but it's a bit clunky to start typing, or swiping up to use the back button.

Thanks guys. Love the app.

loptow:
After testing power tools a bit more, it is super clunky just trying to type stuff, and coming out of immersive mode :s. Temporary solution for sure.

jseybold:
I have also been looking for this feature.

I cloned the repo and took a crack at it by adding this to a few of the classes:

`    @Override
    public void onWindowFocusChanged(final boolean hasFocus) {
        super.onWindowFocusChanged(hasFocus);
        final View decorView = getWindow().getDecorView();
        decorView.setSystemUiVisibility(
                View.SYSTEM_UI_FLAG_LAYOUT_STABLE |
                        View.SYSTEM_UI_FLAG_LAYOUT_HIDE_NAVIGATION |
                        View.SYSTEM_UI_FLAG_LAYOUT_FULLSCREEN |
                        View.SYSTEM_UI_FLAG_HIDE_NAVIGATION |
                        View.SYSTEM_UI_FLAG_FULLSCREEN |
                        View.SYSTEM_UI_FLAG_IMMERSIVE_STICKY);
    }
`

and the following to the onCreate() method of those classes 

`final View decorView = getWindow().getDecorView();
        decorView.setOnSystemUiVisibilityChangeListener(new View.OnSystemUiVisibilityChangeListener() {
            @Override
            public void onSystemUiVisibilityChange(int visibility) {
                decorView.setSystemUiVisibility(
                        View.SYSTEM_UI_FLAG_LAYOUT_STABLE |
                                View.SYSTEM_UI_FLAG_LAYOUT_HIDE_NAVIGATION |
                                View.SYSTEM_UI_FLAG_LAYOUT_FULLSCREEN |
                                View.SYSTEM_UI_FLAG_HIDE_NAVIGATION |
                                View.SYSTEM_UI_FLAG_FULLSCREEN |
                                View.SYSTEM_UI_FLAG_IMMERSIVE_STICKY);
            }
        });
`

[...]

PR #2293, opened 2016-11-19
New feature: Immersive Fullscreen Mode

jseybold:
Hide the status and navigation bars when enabled in Settings > General > Other > Immersive fullscreen mode


jseybold:
Related to Issue #2193

ccrama:
Awesome work, thank you for the PR!

PR #2294, opened 2016-11-19
Fixed restore backup from file

jseybold:
- Mimetype was incorrect, could not select .txt files
- Resolved File Not Found exception when reading/restoring backup file


ccrama:
Good find, and thank you very much for the fixes! Looks like CI is failing, will take a look at that tonight

jseybold:
Glad I could help. It was a useful feature to have so I could run a debug build with my daily configuration.

It looks like there is something wrong with the way the CI server is setup.

ccrama:
Ah you are correct, Travis needs to be set up to accept the new API 25, @Alexendoo  is that something easy to do on your end?

Alexendoo:
Shouldn't be too much of an issue, I'll check it out

PR #2299, opened 2016-11-21
Update build tools

Alexendoo:
Fixes travis build also


ccrama:
Thank you for the fix!

PR #2303, opened 2016-11-23
Add a Share GIF feature in the MediaView activity

gferon:
Here's a PoC PR that enables GIF sharing directly to apps like WhatsApp.

This is merely a proof of concept and not the final version as I have a bunch of questions regarding this feature (and potentially other):
- I wanted to re-use the code between the "Share Image" and "Share GIF" feature but noticed there was a Bitmap to PNG conversion going on: what's the reason behind this? I mostly curious about it, and at first I thought that it'd be a penalty when using mobile and sharing a JPEG image for instance.
- I also wanted to use an horizontal ProgressBar in the BottomSheet instead of progress notification as it seems to be the [recommended material design way](https://material.google.com/components/progress-activity.html) but I saw that you're using https://github.com/soarcn/BottomSheet when a Google implementation that is more complete has been provided starting in the Android Support Library v23.2

Feel free to point me towards any direction that you'd like. I'd be happy to help make bigger changes if you wish.

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Activities/MediaView.java
+++ b/app/src/main/java/me/ccrama/redditslide/Activities/MediaView.java
@@ -21,6 +21,8 @@ import android.os.Bundle;
 import android.os.Environment;
 import android.os.Handler;
 import android.support.v4.app.NotificationCompat;
+import android.support.v4.content.ContextCompat;
+import android.support.v4.content.FileProvider;
 import android.support.v4.view.animation.FastOutSlowInInterpolator;
 import android.text.Html;
 import android.util.Log;
@@ -253,8 +255,8 @@ public class MediaView extends FullScreenActivity
                 if (type.equals("GIFV") && new URL(contentUrl).getHost().equals("i.imgur.com")) {
                     type = "GIF";
                     contentUrl = contentUrl.replace(".gifv", ".gif");
-                    //todo possibly share gifs  b.sheet(9, share, "Share GIF");
                 }
+                b.sheet(9, image, getString(R.string.mediaview_share, type));
             } catch (MalformedURLException e) {
                 e.printStackTrace();
             }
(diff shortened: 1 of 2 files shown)


ccrama:
Thank you for looking into implementing this! 

As for the conversion, that was the old system of saving images and I've been converting the system to use the direct file instead of converting it. The reason the share menu still uses it is I can be sure the image already exists in memory and can share it instantly instead of waiting for a potential download, but if a progress bar was implemented into this system, that would be a non-issue. 


If you would like to look into making a bottom sheet implementation with a seekbar (sort of like how Google Photos does it), I think that would be very slick! Let me know if you have any questions about the code or anything I can point you to.

Nxt3:
@gferon Any updates on this? I'd love to see this implemented.

gferon:
I absolutely forgot about it! I'll try to bring it up and finish it ASAP.

gferon:
Here you go, I cleaned up the code, and tested on my Nexus 6P on Android 8.1
Had to change a bit the way the Intent is built + which context starts the sharing activity, but it finally works!

I think having a ProgressBar in the BottomSheet is out of scope, since this would require to start using the BottomSheet from the Android Support Library - though I'd be willing to port over all the existing usage of `com.cocosw:bottomsheet` in another PR.

ccrama:
Looks great, thank you for the PR! Would you be willing to make the changes that would allow users to share MP4s for any links that aren't gifs? I'd say most of the links are actually videos increasingly the last few months because of v.redd.it. We already save these files in an accessible location, so it should be more straightforward and not require a download! If you would like to help I can point you to the right location :)

gferon:
Sure thing: I suppose I have to look in OpenVRedditTask and make the GIF sharing part more generic / re-using exiting part?

I'll probably also extract the new share AsyncTask in a separate class, since it doesn't really need to access any parent class members.

ccrama:
What I think we could do is actually make a nullable callback that we pass to the GifUtils functions, and we can set that callback to do the actual file sharing if it's an MP4, or do a new gif task if not. I believe I already have some of the framework to be able to do that (the saving gif code is actually a callback), might be more clean to do it that way since the code to actually do anything with the callback would be in the calling class' code and not somewhere in GifUtils. What are your thoughts?

PR #2305, opened 2016-11-27
Immersive fullscreen mode improvements

jseybold:
Checking for focus and acting appropriately seems to fix a few issues with appearance when this feature is enabled. It's not a huge difference, but it seems like a move in the right direction.

When closing the keyboard, before (left) and after (right):

When opening the search dialog from inside a thread, before (left) and after (right):

When selecting the menu option on a post, before (left) and after (right):


jseybold:
I wont be able to check on this for a few days, but I'm not in a rush to have it merged and would welcome testing and comments.

ccrama:
Thank you for this pr! I'll take a look after classes today but it looks like it would solve the main issues people are having with the new immersive mode

jseybold:
The only thing I've noticed so far is the sidebar subreddit search is messed up. If you click the hamburger menu, then select "Go to subreddit", the keyboard pops up but the textbox loses focus so you can't enter anything in it. Disabling immersive mode seems to fix it so I think it's related to something I did.

ccrama:
So I've been trying to look for a solution to the issue to the text field focus not working, do you think we should exit immersive every time a user goes into a text field?

jseybold:
Sure, why not? Immersive mode is great for surfing but if you're actually going to type something out then I don't see a problem with dropping out of immersive mode temporarily.

ccrama:
Sounds good, will do once finals are over (next week). Sorry I've been pretty absent recently

ccrama:
I think everything is working fine! Going to merge after fixing the conflicts and then look into doing the text field thing. The issue with that is we would have to add a focus listener to each text field, as the global listener is triggered by other focus changes and not just the keyboard

jseybold:
The two solutions that I have found so far to address the EditText issue are to:

1. Implement a new class that extends EditText and then change all references in the layout xml to reference that class [(link)](http://stackoverflow.com/a/32290404) or,

2. Add a global listener that checks for instances of EditText and acts accordingly [(link)](http://stackoverflow.com/a/23731329)

I plan to look into these options when I get some free time.

ccrama:
I think solution #1 is the most versatile, it would also allow us to add global features to the EditText if needed in the future. Thank you for taking a look into it! For now, I've updated the settings screen to label "Immersive Mode" as "experimental" so users will not report this as an issue

Issue #2319, opened 2016-12-11
Suggestions for data saving mode


touzainanboku:
Slide version: 5.5.2
Android version: 6.0.1

This is gonna be pretty long, since I tried to be as comprehensive as possible. Sorry about that. 

<h2>How the current data saving mode behaves</h2>

<h3>Testing method</h3>

Slide doesn't show what *exact* link it's loading, and I only have extremely basic programming knowledge, so my method isn't really ideal, but it's the best I could come up with:

- Set data saving mode in Slide, Clear its cache through Settings -> Apps, remove it from recents

- Send a message to myself in Telegram containing a link to a post, and links to different thumbnail/preview versions from reddit/Imgur ([example message](http://pastebin.com/mDUWKHWB))

- Copy the message, and open the post

- Open the image, wait for it to load and then close it

- Paste the message from Telegram in a comment, preview it

- Open every single thumbnail link and see which one loads from cache (and looks the same as the one from the post itself)

My internet is slow enough that I can actually see if it's loading anything from the internet or not. 

As I'm sure you know, reddit has a bunch of previews in the .json file of each post. According to [this](https://github.com/reddit/reddit/compare/2f7837e826cd0be34eabbb1975578545e5f5631c...df671653f2f6b781c08659372be27d765e7598f1#diff-6850f36502467acb3983f3a877321076R583) and the actual .json files, the width of these previews is 108, 216, 320, 640, 960, and 1080. Naturally, they won't be created if the image isn't large enough. As for Imgur, according to [this](https://api.imgur.com/models/image), each thumbnail has a suffix, and they're as following:

suffix | Resolution
-- | -- 
s | 90x90 square
b | 160x160 square
t | 160x160 thumbnail 
m | 320x320 thumbnail
l | 640x640 thumbnail
h | 1024x1024 thumbnail

There also seems to be "r", but I'm not entirely sure what it is, so I didn't include it in this test. 

<h3>Results</h3>

I tested using 5 different posts: an i.redd.it image, an Imgur image, an Imgur album, and two non-Imgur/i.redd.it images. Here are the results:

\# | Post | With reddit data saving | With Imgur data saving 
-- | -- | -- | -- 
1 | [i.redd.it](https://www.reddit.com/r/EarthPorn/comments/5hm3v6) | reddit 640 | reddit 640
2 | [Imgur](https://www.reddit.com/r/EarthPorn/comments/5hkz4y) | Imgur h (1024) | Imgur m (320)
[...]

PR #2326, opened 2016-12-17
More data saving options

touzainanboku:
My attempt at implementing #2319. I'm a real novice when it comes to programming, and it's my first time using GitHub, so I apologize beforehand for any mistakes/issues. 

Basically, this will give the user three different data saving modes:

Low: uses Imgur m and reddit 320
Mid: uses Imgur l and reddit 640
High: uses Imgur h and reddit 1080 (falls back to 960 if 1080 is not available)

For non-Imgur images, when there isn't a preview in the desired size, it's supposed fall back to the full sized preview from reddit (not to be confused with the actual image), since that's closer to the desired resolution than the highest resolution resized reddit preview.

So far I haven't found any issues while testing. I'm really enjoying being able to choose a specific quality, and having it behave more consistently across different images, so I'd really, really appreciate it if you could consider merging this.

Thanks for the awesome app!

(I redid the pull request because the previous one was 4 commits)


ccrama:
Sorry, have been out lately and haven't had time to check the PR out yet. I'll definitely take a look soon, though, thanks for the work!

ccrama:
Checked it out and everything looks good! Thanks for the PR :)

touzainanboku:
Thank you!

PR #2335, opened 2016-12-23
Immersive Fullscreen Mode Fixes

jseybold:
I reintroduced the onSystemUiVisibilityChange listener because it fixes the appearance when selecting and deselecting text fields. It was sensible to remove it given the EditText issue, but I think I have resolved that.

Regarding the EditText problem, it looks like the issue was caused by `descendantFocusability=beforeDescendants` being set in drawer_loggedin.xml. Setting this to `descendantFocusability=afterDescendants` seems to clear the issue right up. I'm not exactly sure why it's being explicitly set to `beforeDescendants` here, but my change doesn't seem to have any adverse effect. With this change, I don't think we need to go through and implement a custom EditText class.

[Reference](http://stackoverflow.com/questions/20406472/edittext-in-listview-loses-focus-when-pressed-on-android-4-x)

Commit Notes:
- Set onWindowFocusChanged check back to if(hasFocus) instead of if(!hasFocus)
- Exiting fullscreen mode on text entry fields should now properly re-enable status/nav bars
- Fixed focus issue on "Go to subreddit" EditText element
- Tested on Google Pixel running v7.1.1 (NMF26Q)

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Activities/BaseActivity.java
+++ b/app/src/main/java/me/ccrama/redditslide/Activities/BaseActivity.java
@@ -57,7 +57,7 @@ public class BaseActivity extends PeekViewActivity
     public void onWindowFocusChanged(final boolean hasFocus) {
         super.onWindowFocusChanged(hasFocus);
         if (SettingValues.immersiveMode) {
-            if (!hasFocus) {
+            if (hasFocus) {
                 hideDecor();
             }
         }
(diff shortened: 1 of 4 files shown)


ccrama:
Sorry for the late reply, took some time off during the holidays. 

Great work, thank you for figuring out how to fix this issue :)! I'll do some testing over the next few days and merge it before the next update. 

Cheers!

jseybold:
No worries, I can just build my own debug version and use that while I wait. I don't imagine that you'll encounter any issues, but I'd be interested to know if you find anything.

Regarding the descendentFocusability, that is odd. I did some research and it appears that fullscreen immersive mode is supported as of version 4.4 [(API level 19](https://developer.android.com/training/system-ui/immersive.html)). Currently we do not have version checks around this feature, but it may be worth implementing them. Maybe we could do an API check wherever that EditText element is referenced and set it dynamically based on API level.
* API >= 19 would be `descendantFocusability=afterDescendants` (immersive mode available)
* API < 19 would be `decendantFocusability=beforeDescendants` (immersive mode unavailable)

That would hopefully solve the issue for versions below and above Kitkat. If the issue impacts Kitkat, we would either have to choose between supporting immersive mode on Kitkat or the EditText field behaving properly unless a workaround could be found that supports both features.

Happy holidays!

ccrama:
Sorry for the late reply, have been working on the iOS version and moving back to school. 

Definitely looks like that fixes the issue, I'll fix that soon and merge your PR. Thank you again for your contributions and looking into this issue!

ccrama:
Will be releasing an update soon with this fix. Thank you very much for your continued contributions!

jseybold:
Just got the update from the play store. It's looking good on my phone. Hopefully other people are getting similar results. Thanks.

ccrama:
I believe they are! Thank you again for the work

Issue #2341, opened 2017-01-01
Does not concider ethernet as form of connection


ioangogo:
Slide version: 5.5.4(F-droid)
Android version: 6.0.1

Due to a overpopulated Wi-Fi spectrum in my area I have resorted to using Ethernet with my phone, I have discovered that slide does not consider it a form of connection.

I have found the code that causes this issue, [This Status enum](https://github.com/ccrama/Slide/blob/master/app/src/main/java/me/ccrama/redditslide/util/NetworkUtil.java#L70-L77) and [this case statement](https://github.com/ccrama/Slide/blob/master/app/src/main/java/me/ccrama/redditslide/util/NetworkUtil.java#L28-L41) as the app assumes it has no connection if the device has no Wi-Fi or mobile. What should be happening is that Ethernet should be accounted for, or other is default if their is a connection and if no interfaces have internet, it should then fallback to none.

ioangogo:
I have a fix that is basicly adding some cases to it that adds the other connection type to their corrersponding network types. I am putting it here as i am struggling to build the app from cli and i am not on a network to get the studio at a decent speed

```
switch (activeNetwork != null ? activeNetwork.getType() : CONST_NO_NETWORK) {
            case ConnectivityManager.TYPE_WIFI:
            case ConnectivityManager.TYPE_ETHERNET:
                return Status.WIFI;
            case ConnectivityManager.TYPE_MOBILE:
            case ConnectivityManager.TYPE_BLUETOOTH:
            case ConnectivityManager.TYPE_WIMAX:
                return Status.MOBILE;
            default:
                return Status.NONE;
        }
```

ioangogo:
Closing this as it has been merged, just waiting for the next update

PR #2342, opened 2017-01-03
Added support for more connection types

ioangogo:
This is the fix found in Issue #2341 seems to work on my phone so far


ccrama:
That's an interesting issue, thank you for the PR and the fix! Works great

PR #2345, opened 2017-01-06
Android automatic refactor - Recycle

luiscruz:
Hi,

This pull request has the changes generated while applying the rule "Recycle".

Some resources (e.g., ```Cursor``` instances) should be closed when they are no longer necessary. 

Please consider them and let me know if you agree with them.

Best,
Luis


bb010g:
This should be refactored to use a try-with statement, to eliminate the null check & be visually cleaner.

l3d00m:
@bb010g Try with ressources requires minSdkVersion 19. Slide is currently using minSdkVersion 15 though, so it has to use a manual try catch.

Tensounder54:
@luiscruz Please resove the conflics in the branches. These must be resolved before we will consider your pull request.

Have a great day.

> ↳T*54*

PR #2346, opened 2017-01-06
Android automatic refactor - ViewHolder

luiscruz:
Hi (again),

This pull request has the changes generated while applying the rule "ViewHolder".

When developing list views, it is very common to repeat work in every item. Something very common is retrieving a field with the API method ```findViewById```. There is a ViewHolder pattern that lets you reuse the work made in the previous computed item.

Please consider them and let me know if you agree with them.

Best,
Luis

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Adapters/SubChooseAdapter.java
+++ b/app/src/main/java/me/ccrama/redditslide/Adapters/SubChooseAdapter.java
@@ -69,6 +69,10 @@ public class SubChooseAdapter extends ArrayAdapter<String> {
 
     private static class ViewHolderItem {
 		private TextView t;
+
+        ViewHolderItem(TextView t){
+            this.t = t;
+        }
 	}
 
 	@Override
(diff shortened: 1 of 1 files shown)


app/src/main/java/me/ccrama/redditslide/Adapters/SubChooseAdapter.java
```diff
@@ -67,12 +67,23 @@ public Filter getFilter() {
...
 
-        final TextView t =
-                ((TextView) convertView.findViewById(R.id.name));
+	@Override
+    public View getView(final int position, View convertView, ViewGroup parent) {
+        ViewHolderItem viewHolderItem;
+		if (convertView == null) {
+			convertView = LayoutInflater.from(getContext()).inflate(R.layout.subforsublist, parent, false);
+			viewHolderItem = new ViewHolderItem();
+			viewHolderItem.t = ((TextView) convertView.findViewById(R.id.name));
```
bb010g: This should be refactored into the `ViewHolderItem` constructor call, not a separate assignment.

ccrama:
Thank you for the PRs! Looks great, merging now.

PR #2347, opened 2017-01-06
Android automatic refactor - ObsoleteLayoutParam

luiscruz:
Hi,

This pull request has the changes generated while applying the rule "ObsoleteLayoutParam".

While developing your application's views you might be specifying attributes in a view's artefact that are not necessary due to the nature of its parent. In this PR, those attributes were replaced by a comment.

Unfortunately, this tool is not able keep the original whitespace of the files, so comparison without ignoring whitespace might be confusing.
Please consider the changes and let me know if you agree with them.

Best,
Luis


ccrama:
Thank you again for this PR Luis!

The layout does look correct, but I will do some more testing before a merge. 

Definitely looks to have simplified the views, though, so thank you for the refactoring!

luiscruz:
Ok, I'll do that ;)

PR #2353, opened 2017-01-11
Allowed users to search through subreddits even when offline

yalmadani:


ccrama:
Looks great! Thank you for the PR :)

PR #2372, opened 2017-02-15
Fix crash when selecting some items in Mod Log

### LINKED ISSUE #2338: [Bug] Mod Tools - Selecting "Ban User" and "Wiki Revise" crashes Slide
Slide version: 5.5.4
Android version: 7.1.1

Selecting either "Ban User" or "Wiki Revise" within Mod Tools' Mod Log crashes Slide.

As a side note - didn't want to make another issue about it but can if needed - could the Ban User entry list the banned user's username?

l3d00m:
fixes #2338


ccrama:
Good catch, thank you for the fix!

PR #2391, opened 2017-03-16
Remove min 21 config

Jawnnypoo:
This is no longer needed, yay!
https://twitter.com/danlew42/status/838799239576612865


ccrama:
Woo! Really happy that happened, don't have to keep changing the min version number down every time I want to debug install to my KitKat devices

Jawnnypoo:
Yes, exactly 😄

PR #2392, opened 2017-03-16
Update many dependencies

Jawnnypoo:
Pretty straightforward. Small update to make ProcessPhoenix a bit more stable

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -73,57 +73,64 @@ android {
 
 }
 
+ext {
+    supportLibVersion = '25.3.0'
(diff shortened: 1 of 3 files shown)


ccrama:
Looks great, thank you very much for the PR!

File definitely is a lot cleaner (thanks for making the support versions a var), and thanks for the little ProcessPhoenix fix. 

Just because so many repos were updated, I will probably release this next update as an alpha first to make sure everything is still working well together

Jawnnypoo:
No worries, that makes sense to me. Happy to help out!

PR #2417, opened 2017-04-20
Adjusted tests to work with and test pinned subreddits

freundTech:
https://github.com/ccrama/Slide/commit/75efd3cf91ad0a18c69d9890660063d3ba565e04#commitcomment-21828957

This would be an alternative to adding an extra parameter to fix the tests. It has the advantage that instead of just ignoring pinned subreddits it also tests them.


Alexendoo:
All good 👍

PR #2425, opened 2017-04-24
Added support for opening mailto: links externally

tryton-vanmeer:
For #2420.

ContentType.getContentType returns EXTERNAL for urls that start with 'mailto:'.
When LinkUtil.formatURL checks if a url doesn't contain '://', it also makes sure the url doesn't start with 'mailto'.

There still exists an issue with markdown links for mailto as described in this issue -> reddit/snudown#67


ccrama:
Awesome, thank you for the PR! I'll fix the conflicts and merge before the next update

PR #2431, opened 2017-04-27
Pinned subs toolbar fixes

Alexendoo:
Allows you to select multiple pinned subreddits and unpin them in one action, and swaps the lock icon out for a pin in the toolbar


ccrama:
Awesome, thank you for the PR! Looks great

Issue #2430, opened 2017-04-26
Collapsing comments


alde:
Slide version: 5.6
Android version: 7.0

This is either a feature request, or a question.
Is there a way to collapse the entire comment thread when I'm done with it (for example like swiping left on a thread in Alien Blue on iOS). The way I enjoy redditing is collapsing the threads I'm done with, and I'd prefer to not have to scroll back up to the parent to collapse it just so I can move on.

ccrama:
Thank you for the suggestion! That is an interesting idea, but how does that differ than using the comment navigation arrows to jump down to the next parent comment chain?

alde:
Ah, did not know about that.

In alien blue, coming back to a comment section would still have collapsed threads collapsed, showing that I consider them done.

But the navigation arrows covers most of my use-case, so thanks for that.

Scrumplex:
This issue can be closed as this has been implemented (for a long time?).

PR #2433, opened 2017-04-30
Add setting to collapse parents

alde:
This PR adds a setting to collapse parents when collapsing a child comment.

A proposed solution to #2430.

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Activities/SettingsComments.java
+++ b/app/src/main/java/me/ccrama/redditslide/Activities/SettingsComments.java
@@ -120,6 +120,17 @@ public class SettingsComments extends BaseActivityAnim {
             });
         }
         {
+            SwitchCompat single = (SwitchCompat) findViewById(R.id.collapse_parents);
+            single.setChecked(SettingValues.collapseParents);
+            single.setOnCheckedChangeListener(new CompoundButton.OnCheckedChangeListener() {
+                @Override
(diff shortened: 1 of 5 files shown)


ccrama:
Really interesting, thank you for the PR! 

One thing I noticed is that you change the ViewHolder and the comment source on the doOnClick function, which also handles hiding the menu if the menu is currently displayed on a comment. It appears that it would hide the top-level comment's menu when running this code, and a better solution might be to only change the source of the comment for the hideAll() function.

alde:
Ah, I guessed that there would be a more logical place, but I had trouble finding it. It's my first time poking around with android, so trying to follow the flow of the code was not obvious to me.

I'll see if I can come up with a better solution.

alde:
Ah now I remember, I tried editing only hideAll(), but the problem I ran into was that it made me have to click twice, once to collapse the children, and once to collapse the parent.

I have done a small refactoring of the doOnClick to better isolate the side-effects of the setting, and as far as I have been able to test I haven't found any issue

ccrama:
Thank you for looking into it! 

That is an interesting side effect, have you found any other side-effects of your updated code? And sorry for the size of that class, it is one of the most complicated classes to work with so thank you for giving it a shot!

alde:
Yes, the counter of hidden items f
would be children of the clicked comment, not actual hidden comments

But with the refactored code things seem to work as expected.

PR #2473, opened 2017-06-27
fix crash when syncing subreddits while user is not logged in

YusefOuda:
when the user is not logged in and tries to sync subreddits (settings -> Manage your subreddits -> Sync subscriptions button), the app will currently crash because of an unhandled exception. This PR solves it by catching the additional exception thrown by MultiRedditManager.java


ccrama:
Looks great, thank you for the PR!

Issue #2464, opened 2017-06-01
Slide wants me to rate it on Google Play... I don't have Google Play


Matttter:
Slide version: 5.6.2
Android version: irrelevant

Every once in a while a message pops up in the bottom asking to rate it. When I click it it just opens the app page in F-Droid. It would be nice if there were some kind of check for that.

ccrama:
@YusefOuda  took care of this one, thank you for the report!

PR #2474, opened 2017-06-27
only ask user to rate the app if not f-droid

YusefOuda:
This fixes issue #2464


app/src/main/java/me/ccrama/redditslide/Activities/MainActivity.java
```diff
@@ -1354,7 +1355,7 @@ protected void onPreExecute() {
         }
 
 
-        if (Authentication.isLoggedIn && NetworkUtil.isConnected(MainActivity.this))
+        if (Authentication.isLoggedIn && NetworkUtil.isConnected(MainActivity.this) && !FDroid.isFDroid)
```
bb010g: Just stick the `FDroid` check first so we don't hit `isConnected` when we don't need to.

YusefOuda: good call, made that change @bb010g

ccrama:
Good catch, thank you for the fix!

PR #2481, opened 2017-07-05
add setting to allow single column view in multi-window mode

YusefOuda:
This pull request aims to add a new feature to Slide. This feature will add a new setting to the multi-column settings window called "Single column in multi-window mode". It is disabled by default. When the setting is enabled and the app is in multi window mode, only a single column for submissionview, gallery, search, and multireddits should appear. The setting should have no effect if the app is not in multi-window mode. 

The reason this came up is I like to have dual columns in landscape mode. However, when using the app in multi-window mode, the app still considers itself in landscape mode (even if the phone is held in portrait). This happens because the width is greater than the height. This makes it hard to read the submissions as they are shown in dual column mode on a very small width. This setting should make it much easier to browse when using this configuration

Let me know if you have any questions.

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Activities/Gallery.java
+++ b/app/src/main/java/me/ccrama/redditslide/Activities/Gallery.java
@@ -1,6 +1,8 @@
 package me.ccrama.redditslide.Activities;
 
+import android.app.Activity;
 import android.content.res.Configuration;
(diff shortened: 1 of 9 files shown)


ccrama:
Really great stuff, thank you for the PR! Noticed this behavior as well, this is a much better solution to that

Issue #2468, opened 2017-06-04
Be able to share content via redd.it short links


KeronCyst:
Slide version: 6.5,2
Android version: 7.1.2

At least, whenever a post has one. It'd be nice on the clipboard, haha.

PR #2484, opened 2017-07-09
share reddit shortlink when possible

YusefOuda:
addresses issue #2468. I tried to catch any instances where links are being shared and change it over to the shortlink. I wasn't able to test each code path since I wasn't able to pinpoint where each file was being used in the app. Any insight would be appreciated. Let me know if you see any issues

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Adapters/ContributionAdapter.java
+++ b/app/src/main/java/me/ccrama/redditslide/Adapters/ContributionAdapter.java
@@ -259,7 +259,7 @@ public class ContributionAdapter extends RecyclerView.Adapter<RecyclerView.ViewH
                         @Override
                         public void onClick(View v) {
                             if (submission.isSelfPost())
-                                Reddit.defaultShareText("", "https://reddit.com" + submission.getPermalink(), mContext);
+                                Reddit.defaultShareText("", "https://redd.it/" + submission.getId(), mContext);
                             else {
                                 new BottomSheet.Builder(mContext)
                                         .title(R.string.submission_share_title)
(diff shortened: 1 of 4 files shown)


YusefOuda:
@ccrama any input on this PR and also https://github.com/ccrama/Slide/pull/2485? much appreciated

bb010g:
I don't know how I feel about this. Given the choice between sharing the permalink and a shortened URL, I'd normally go with a permalink. Could this be exposed behind a preference to use shortened by default, or some other way that provides choice when sharing?

Nxt3:
I think, if anything, this should be a setting under "Link Handling". I personally don't care either way, but I would imagine there are people who do. Making it a toggle would be better than asking the user every single time they want to share a link.

ccrama:
I agree with @Nxt3 on this one, shouldn't be too difficult to add a setting in for it! If you would like me to do that @YusefOuda let me know. Also sorry for being absent on Github, I have been primarily focused on Slide iOS which went into alpha testing yesterday, so things are a little busy on my end!

YusefOuda:
I like the idea of a setting, that makes sense. I can take a stab at this tomorrow

ccrama:
Hey just wondering what the status of this was? I can implement the setting if you'd like

YusefOuda:
That's a good question. It's been a while, I remember being unhappy with some aspect of what I did. It's not coming back to me what that was though

ccrama:
We were just going to make a setting for it so users can disable the shortlink if they want to. I think other than that it was working great!

YusefOuda:
Cool, yeah if it looks good to you feel free to add the setting and we can get this merged in

ccrama:
Decided to merge it without an option for now. If you're still interested in an option, feel free to make an issue @bb010g  and I'll take another look at it

Issue #2475, opened 2017-06-29
Ask for confirmation before uploading to imgur


haadr:
Slide version: 5.6.2
Android version: 7.1

Hi all and thanks for a great reddit reader.

I just accidentally uploaded an image to imgur (apparently?) via the share functionality. When sharing an image to Slide in android, it seems to immediately upload it to imgur. Say I'm clumsy or somehow manage to accidentally upload a private image to imgur, that would be very very bad. Currently there seems to be no way to abort this procedure. I strongly propose at least requiring a confirmation button before anything is immediately uploaded to the internet for potentially public viewing. The share button is very prominent when viewing images in the gallery, and then it's just another accidental tap to upload whatever to the internet.

Regards,
haadr

Nxt3:
Someone correct me if I'm wrong, but the way Slide uploads to Imgur, the images aren't public. So even if you did, nobody can freely see them. I might be wrong though.

haadr:
If that's true (and it would make sense) that's, good. They're still uploaded to imgur though, who will have them, and they still wouldn't be private per se. Since this is an action that is arguably impossible to undo (depending on how paranoid you wanna get), I still think it should require prior confirmation.

ccrama:
Done in the latest alphas

MustardCat:
Did this change get reverted?  I just recently uploaded an image to imgur and wasn't prompted for confirmation (v5.7.1)

ccrama:
@MustardCat  this is only on the Submit screen

PR #2485, opened 2017-07-10
ask for confirmation before uploading to imgur

YusefOuda:
addresses issue #2475 

Adds another MaterialDialog which will ask the user for confirmation before starting the upload. On confirmation, will execute the AsyncTask


ccrama:
Might save some people, thanks for the PR!

PR #2486, opened 2017-07-12
fixes notification action icons

YusefOuda:
fixes scaling issue as described [here](https://redd.it/6muloj). Icons from https://material.io/icons/


ccrama:
Thank you for the fix! Was the issue just that the icon was too large for the display density?

YusefOuda:
yup, that seemed to be it

PR #2502, opened 2017-08-04
[Do not merge] Fetch comments concurrently in a thread pool

AdamNiederer:
Hi there,

I've been trying to improve the speed of offline subbreddit caching, and I figured that serially awaiting each HTTP request was eating the bulk of the time.

While fixing that problem, I found that `RedditClient.execute()` has some kind of lock which doesn't let me query Reddit concurrently. Although this patch doesn't hurt the wall time of the caching process, I'd recommend not merging this until I can figure out how to get around that limitation. Does reddit have a particularly oppressive rate limiter or something?

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/CommentCacheAsync.java
+++ b/app/src/main/java/me/ccrama/redditslide/CommentCacheAsync.java
@@ -29,8 +29,13 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.BlockingQueue;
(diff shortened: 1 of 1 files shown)


ccrama:
Thank you for the PR! 

Actually, this is done on purpose because Reddit has API rate limitations, and the caching mechanism was designed to not break these limits. Unfortunately don't know if your changes break these limits, but I know that JRAW has a master rate limit built into it (I figured serially downloading would ensure we don't break any limits on our side either). One way we could do this is to execute each as a separate pooled thread, but keeping track of all of those threads is pretty messy and that's why I opted to just have one master async thread. If you have any other suggestions on how to go about this, I'm open to any suggestions!

PR #2509, opened 2017-08-07
fixes for accent color array being too big

YusefOuda:
Fixes issue described here: https://www.reddit.com/r/slideforreddit/comments/6rjre4/bug_app_crashes_when_you_select_a_unsupported/


ccrama:
Huh, I think the issue here is actually that there isn't a theme value for a certain color. What base themes did this affect?

YusefOuda:
This issue affected every base theme as far as I could tell. The last value in the array was always empty, but it was still selectable in the UI causing it to crash. Do you think the `ColorPreferences.Theme` enum is missing some colors then?

Regardless, I think the helper method I created can stay right? Dividing the number of total values by 7 seemed kind of arbitrary to me, but maybe there was something I was missing

ccrama:
True, yes your fix is a better solution either way. I do believe some of the enums are missing, will investigate it myself! Thanks again for the PR

PR #2530, opened 2017-09-19
Fixed failing build

freundTech:
This fixes build failing with API level 26, by resolving a naming conflict (attr/font) and updating robototextview. 
In addition robototextview is switched back from the fork to the upstream version
and the gradle plugin is used to exclude unneeded fonts

Sadly this now crashed when starting the app, but at least it's better than not building at all.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -3,10 +3,11 @@ plugins {
 }
 
 apply plugin: 'com.android.application'
+apply plugin: 'com.devspark.robototextview.gradle-plugin'
 
 android {
-    compileSdkVersion 25
-    buildToolsVersion '25.0.2'
+    compileSdkVersion 26
+    buildToolsVersion '26.0.1'
     lintOptions {
         quiet true
         abortOnError false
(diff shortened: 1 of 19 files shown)


ccrama:
Thanks for the PR! Glad to see you around :)

Unfortunately I have run into many more issues updating to API 26, and although this builds, it is an instant crash on all devices. I have done a lot of test fixes on my local repo, and am trying to get it to a building and running state asap. I may copy parts of this PR into my commit, but I believe your changes conflict with the current ones locally. Thank you for fixing the build part! 

The main issue currently is GSON has Java 1.8 code that can't be converted down to 1.7, so I have downgraded to an older GSON version to fix that conflict. Additionally, it seems as if there are missing or conflicting AppCompat files that cause crashes, and I'm looking into a fix for that. N was a super easy upgrade, wasn't expecting O to break this much :/

ccrama:
Found the culprit! Have merged your changes into my local and will push this tonight sometime :)

PR #2536, opened 2017-09-21
Fixed robototextview including too many fonts

freundTech:
Important: This needs a complete rebuild in order to apply. The robototextview gradle plugin doesn't recognize changes to it's config and will continue including the old fonts without a full rebuild!

@ccrama


ccrama:
Great find, will do a full rebuild after merging this! Thanks for the PR

PR #2537, opened 2017-09-22
Update OpenRedditLink.java

garretyoder:
Just happened to notice this while browsing.


ccrama:
Good catch, thanks for the fix!

PR #2541, opened 2017-09-22
Fixed webview crash and broken logic.

### LINKED ISSUE #2539: Opening a link within a comment crashes Slide
Slide version: 5.6.5-alpha5
Android version: 7.1

Link to comment, then click wallpaper link: https://reddit.com/r/androidthemes/comments/71amkm/theme_tranquility/dn99rgc?context=3

freundTech:
WebView crashes when accessed inside the WebViewClient.

Also a not was missing.

Fixes #2539/#2540


ccrama:
Thank you for the PR! I actually just fixed it on master before I saw this PR :( 

The URL check needs to compare to the currently loaded website and not the URL it is checking from (if I were on twitter.com and there was a tracking URL, I would need to let that go through). 

Either way, is fixed now. Cheers!

Issue #2534, opened 2017-09-21
Blocking 


EboniLM:
Slide version: 5.6.5.1
Android version: 7.0

Any plans to implement in-app blocking? Ha ing to do it through the desktop website is tedious.

l3d00m:
It's in the Settings and it's called "Filters", if by blocking you mean to filter based on keywords, title or subreddit.

Nxt3:
I think he means blocking other users?

EboniLM:
Yes, i do mean blocking other users.

ccrama:
Added by @YusefOuda !

PR #2558, opened 2017-10-03
adds ability to block user from profile

YusefOuda:
Addresses issue #2534

JRAW did not have this ability, so had to add it in. Let me know what you think. Doesn't look like there is a way to unblock through the API, only block.


ccrama:
Looks great! Thanks for the PR! I'm going to extend this to work from the inbox as well if you want to block someone from the PM menu

PR #2564, opened 2017-10-14
resolves #2489

### LINKED ISSUE #2489: Notification slider for message unread don't update
Set the app in French.
Go on message -> menu -> Notifications.

You have a checkbox "Recherchez de nouveaux messages chaque 6 heures"
An other checkbox.
A slider.

Change the slider and the text of the first checkbox change and skip the time indication: "Rafraichir les messages toutes les" [time info is missing here]

1) When changing the slider I want to have the info of the selected time.
2) Both string are different, (look like the string when unselected)
3) What happen if I uncheck (never check or check when I fetch a subreddit?
4) I think both checkbox should be inverted: the slider is too far away and still visible when the first checkbox is unselected. The second checkbox have no need to be in the middle.

Slide version: 5.6.2
Android version: 7.1.1

Zykino:
I don't know how to find the 20 strings that are missing

### DIFF
--- a/app/src/main/res/values-fr-rFR/strings.xml
+++ b/app/src/main/res/values-fr-rFR/strings.xml
@@ -491,7 +491,7 @@
   <string name="settings_mail_check">Rafraîchir les messages</string>
   <string name="settings_mail_info">Plus le rafraîchissement est fréquent, plus Slide risque de consommer de la batterie en arrière plan !</string>
   <string name="settings_main_color_title">Modifier la couleur principale</string>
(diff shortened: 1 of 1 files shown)


Alexendoo:
Hi, thanks for your contribution, we use crowdin to manage translations so would you be able to fill it in there? https://crowdin.com/project/slide-for-reddit

It should also show you any missing strings

Zykino:
Ok, done. didn't saw a contributing.md so didn't care about reading again a readme, sorry.

Alexendoo:
No worries, thanks for the help @Zykino

Issue #2577, opened 2017-10-26
Enable Wide Color Gamut support when viewing media


Nxt3:
Need to add `setColorMode()` to the following Activities:

- Album
- AlbumPager
- MediaView
- Tumblr (?)
- TumblrPager (?)

Make sure to add in the necessary checks for if a device even supports a wide color gamut and if the device is running Android 8.0+.

- [setColorMode()](https://developer.android.com/reference/android/view/Window.html#setColorMode(int))
- [isScreenWideColorGamut()](https://developer.android.com/reference/android/content/res/Configuration.html#isScreenWideColorGamut())

PR #2579, opened 2017-10-29
Added wide color gamut to media activities

Nxt3:
Addresses #2577 

@ccrama Let me know if I missed any activities that display media.


ccrama:
Thanks for the PR! What does that exactly mean and what are the effects of doing this?

Nxt3:
On devices running 8.0+ that have capable screens, the color mode changed to `wideColorGamut`. On my Pixel 2 XL, this means that when I'm viewing media in Slide, colors are boosted and become more lifelike.

As far as I know, the only downside is battery life; that's why I'm only doing it for media viewers and not the entire app. Even then, battery life shouldn't be affected much.

PR #2588, opened 2017-11-09
sort pinned subreddits when dragged

YusefOuda:
Currently, when sorting subreddits alphabetically, the pinned subreddits are sorted by when they were added to the pinned list. This pull request will update the sort of the pinned list every time one of the pinned subreddits is moved. This will help keep the pinned subreddits in the order the user wants when sorting the list alphabetically.


ccrama:
Really good addition! Thank you for the PR :

PR #2592, opened 2017-11-18
New gifs

ccrama:
This removes most Gfycat conversion for gifs and instead relies on Reddit to get gifs in submissions, and uses a web display for gifs in comments

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ android:
   components:
   - tools
   - platform-tools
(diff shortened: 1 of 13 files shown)


ccrama:
This build also updates to the latest beta of gradle, which is causing the Travis issues

Alexendoo:
It was because of the build tools, that should do it. I moved the gradle plugin onto the stable version now 3 is out properly also

PR #2594, opened 2017-11-23
Improve seeking in MediaVideoView

sommd:
Use the new seekTo(long, int) on API>=26. This gives finer control over the frame to seek to. On non low-ram devices, SEEK_CLOSEST is used to seek to the nearest frame. On low-ram devices, SEEK_CLOSEST_SYNC is used to seek to the nearest key-frame. This is an improvement over seekTo(long) (which uses SEEK_PREVIOUS_SYNC) in both cases.

Tested on Pixel 2 (8.0.0) and AVD (8.1.0).


ccrama:
Great find, thank you for the PR! Works great and fixes a lot of gripes people were having with the seek bar

PR #2617, opened 2018-01-05
use android-27 in .travis.yml

Alexendoo:
fixes the build failure (this license thing is a real pain)


ccrama:
Good find thanks!

Issue #2615, opened 2018-01-03
External website link gets modified when being opened


magnus-gross:
When I visit [this reddit post](https://www.reddit.com/r/linux/comments/7nqhe8/initial_benchmarks_of_the_performance_impact/) with Slide, and try to open the link with the external browser (Chrome) or with inline view chrome, the link gets changed from `https://www.phoronix.com/scan.php?page=article&item=linux-415-x86pti&num=1` (correct) to `https://www.phoronix.com/scan.php?page=article&amp;item=linux-415-x86pti&amp;num=1` (not available). I noticed that this is like this with every phoronix link.

The expected behaviour is to not modify the URL so that I can view the website with the browser.

Slide version: 5.7.2
Android version: 6.0.1

ccrama:
Slide actually does not modify the URL, it is just being HTML escaped through Reddit's JSON. Where are you tapping to open the link?

magnus-gross:
Woah, that was a fast response :O

I tap on the thumbnail. I noticed that if I explicitly press on "Open Externally", it will open the correct URL. It will also share the correct URL, if I click on "Share Link"

ccrama:
Haha we have a bot on the IRC channel that pings with new issues. That shouldn't be happening as Slide forces an un-escape, but I'll look into it!

PiwwowPants:
Pressing "Open Externally" runs through a different code path than clicking the thumbnail and the app open externally based on default settings.
The former passes through `openExternally(String url, Context context, Boolean encoded)` which does not perform a `StringEscapeUtils.unescapeHtml4` on the url.
The latter passes through `defaultShare(String url, Context c)` which does.

You'd potentially want to consolidate this logic to avoid opening externally in two places.

PR #2632, opened 2018-01-27
Bugfix/unescape all links leaving slide

PiwwowPants:
Fixes https://github.com/ccrama/Slide/issues/2615

I made this change fairly quickly, but I think I tested any cases I needed to (I made sure all link handling modes still function correctly). I do not believe there is an issue running `unescapeHtml `on a url that does not need it or already had it ran. The only issue is if that method has considerable overhead, which I also do not think it does for the benefit of always having URLs handled well.

- The change got rid of the, what was essentially, duplicate `Reddit.defaultShare `method in exchange for using the `LinkUtil.openExternally` method. If there is a reason to not do this, let me know, but it seemed like the functionality was the same and internal links are still handled fine.
- Additionally, copying text was moved out to a `LinkUtil.copyUrl` helper method to keep the logic consistent and `unescape `the html in the urls for copying.
- Custom tabs were also not always correctly unescaping the HTML. I moved the logic for opening a custom tab out to a `LinkUtil.openCustomTab` method and forced it to always unescape the HTML if a custom tab can be opened (prevents doing it twice if no custom tab can be opened)

Finally, every commit is an opportunity to run "optimize imports", so I did.

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Activities/Album.java
+++ b/app/src/main/java/me/ccrama/redditslide/Activities/Album.java
@@ -1,15 +1,10 @@
 package me.ccrama.redditslide.Activities;
 
 import android.app.Activity;
-import android.app.Notification;
-import android.app.NotificationManager;
-import android.app.PendingIntent;
 import android.content.DialogInterface;
 import android.content.Intent;
 import android.graphics.Bitmap;
 import android.graphics.Color;
-import android.media.MediaScannerConnection;
-import android.net.Uri;
 import android.os.AsyncTask;
 import android.os.Bundle;
 import android.os.Environment;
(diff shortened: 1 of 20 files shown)


PiwwowPants:
Oh god, I'm a bad person. Sorry about that.

Alexendoo:
I'm not entirely sure why it has started giving an error about the SDK license, I'll look into it

Alexendoo:
On a separate note, your email address is not associated with your GitHub account (or the commits are using a different email address), so they don't show up as coming from your account currently

https://help.github.com/articles/setting-your-commit-email-address-on-github/

ccrama:
Great cleanup, thank you for merging those methods into the LinkUtils class too! Not surprised the code was copied in a few places, don't know why that method was so badly named in the Reddit class haha

PR #2633, opened 2018-01-27
Workaround travis build failure

Alexendoo:
I'll keep an eye on the thread and remove it when it's no longer needed


ccrama:
Thanks @Alexendoo !

PR #2634, opened 2018-01-27
Replaced usages of legacy "file://" logic for intents with FileProvider

PiwwowPants:
I see no existing issue for this defect.

On Android N+, attempting to view the backup file after creating it causes a crash. This is due to a change in the file handling logic for "file://" that makes it no longer function for these purposes.

- This will resolve any crashing that occurs when attempting to read-write files from locations outside of Slide, including the above
- Additionally this resolves a defect where the camera would crash the app when opened for submission due to a defect in TedBottomPicker
     - https://github.com/ParkSangGwon/TedBottomPicker/issues/47

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -20,10 +20,9 @@
             android:resizeableActivity="true"
             android:roundIcon="@mipmap/icon_round"
             android:theme="@style/yellow_dark">
-
         <provider
                 android:name="android.support.v4.content.FileProvider"
-                android:authorities="${applicationId}.MediaView"
+                android:authorities="${applicationId}.provider"
                 android:exported="false"
                 android:grantUriPermissions="true">
             <meta-data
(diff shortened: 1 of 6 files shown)


PiwwowPants:
And also sorry to be pushing another branch while Travis CI will likely fail, but I had two queued up already.

Alexendoo:
Don't worry about the travis thing, seems to an upstream issue, #2633 takes care of it for now

ccrama:
I think Alexendoo's changes fixed Travis for now, thank you very much for the fixes! Looks like a better way to do the intents, will test this out and get it merged before the final 5.8 alpha

PiwwowPants:
I am fairly sure not all file:// actions were caught. Additionally there are likely (read, are) instances where the file provider is used directly, but since this was really just a single issue fix I didn't want to touch too much more code than I did.

TedBottomPicker is a real problem child I plan on making some changes to.

ccrama:
Merged it from my testing branch, everything works great! Thank you for the pr :)

ccrama:
I will make sure there are no left over file:// actions, but I believe you got all of the uses of file provider to my memory... And if you want to make changes to how Ted works I would be totally cool with it! It has been difficult to implement it fully

Issue #2436, opened 2017-05-03
Night mode settings are 1 hour too early.


richiksc:
Slide version: 5.6.1
Android version: 7.1.2

I set it to 6:00 but it ends at 7:00.

I think this has to do with daylight savings time.

Nxt3:
Is this still an issue?

richiksc:
@Nxt3 I can confirm that the issue still persists.

PiwwowPants:
I am just asking, because maybe my boolean logic is bad, but it might be as simple as a bad conditional in the isNight method.

`return (hour >= nightStart + 12 || hour <= nightEnd) && tabletUI && nightMode;`

So, if the current hour is less than OR equal to the time you set to end night mode, it remains night mode. I am pretty sure dropping that OR equal will fix it. I want it to end when that hour hits.

Does that sound right?

PR #2635, opened 2018-01-28
Removed OR equals operator from hour to nightEnd check in isNight method

PiwwowPants:
Fixes https://github.com/ccrama/Slide/issues/2436
Night mode ends an hour late.

Get ready for a wild code review.
As I mentioned in my comment on the issue, I am not 100% convinced my boolean logic is right, but I am fairly sure it is.
In this "`isNight`" check, we should not be deciding it's night time if the current `hour` is equal to the `nightEnd` time, because that would indicate night mode remains on for the duration of the `nightEnd` hour.

I just removed the = operator off of the nightEnd check and we should be all fixed.

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/SettingValues.java
+++ b/app/src/main/java/me/ccrama/redditslide/SettingValues.java
@@ -458,7 +458,7 @@ public class SettingValues {
 
     public static boolean isNight() {
         int hour = Calendar.getInstance().get(Calendar.HOUR_OF_DAY);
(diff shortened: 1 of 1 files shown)


ccrama:
Thanks for the fix! I think that's exactly what is going on because a 5:00 end time at an hour = 5 would result to true, when the highest hour should actually be 4 (hour < 5)

Also thanks for the recent PRs 👍 will be looking through them the next few days for inclusion in the next alpha. If you ever want to chat with us, we are usually chilling on the Slide IRC channel!

Issue #2455, opened 2017-05-23
Rtl comments


bennypowers:
Please take a look at rtl language comment layouts e.g. r/Israel

Thanks. Great app.

PR #2636, opened 2018-01-29
Changed comment textview to match parent width

PiwwowPants:
Fixes https://github.com/ccrama/Slide/issues/2455

Another wild, single line change.
Comment text was wrapping content, which works fine for ltr content. When rtl content was displayed, it wrapped as if it was ltr, so it would display in the center of the comment text view much of the time.

This change expands the comment text view to be full width so rtl content displays correctly, starting at the very right side of the text view.

How RTL displays before change

How RTL displays after change


ccrama:
Great find, thank you for fixing this bug!

PR #2637, opened 2018-01-29
Removed code path that allowed "Removed by true" to display to mods

PiwwowPants:
No github issue.
Fixes [issue from reddit](https://www.reddit.com/r/slideforreddit/comments/7sgm7d/when_moderating_what_does_removed_by_true_mean/)

Two code paths exist that create the removed line (I literally have no clue what that is). Both do the exact same thing but one does not verify the "removedBy" text is not "true". Per the reported issue on the subreddit, "removed by true" is displaying to the mod.
Presumably, according to the comment, we should just fallback to the "Removed by Reddit" text if this occurs.

I removed the duplicate code path, leaving the one in the helper class (I did the same for createApprovedLine... whatever that is).

I clearly did not test this but it compiles so it's probably squeaky clean.


ccrama:
Thank you for the PR! The createRemovedLine creates a SpannableString that is appended to the title if the user is a mod and if the post has been removed. I guess I created two copies for both submissions and comments, and fixed the bug for comments but not submissions. Your fix makes both submissions and comments use the same method and therefore the comment fix works for submissions too :)

Issue #2397, opened 2017-03-26
Overriding back button behavior from searches


Nxt3:
If you do a Google Search, and it's a Reddit link, the link will open in Slide. However, once you're done and press the back button (and you had Slide open previously), it will take you back to where you were in Slide. Pressing the back button should take you back to Google Search.

ccrama:
Do you have any idea of how to go about that? It seems we would have to make a second instance of Slide

Nxt3:
I think we would create a new app instance, yes; if the user swiped to get out of that post, then it would return back to the Google Search results. This is how Flamingo for Twitter handles it and other apps.

Nxt3:
Let me ask on /r/androiddev and I'll get back to you on the best way to handle this. Could you point me to the file that handles this though? Maybe I can fix it myself.

Nxt3:
Check out my comment: https://reddit.com/r/androiddev/comments/636cwp/weekly_questions_thread_april_03_2017/dfypbap?context=3

ccrama:
I added both of those flags, but it still does the same thing, unfortunately... Wonder if there are any other open source apps that offer this?

PiwwowPants:
I liked the idea so I put together how the change should work in the reference pull request.
I do worry that it may be confusing to a user though. If a user launches Slide via a third party app with an intent and then browses for a long time, this change I made will cause them to completely lose their place when the app is backgrounded.

Let me know how you feel. This seems to be "the way" to get the behavior @Nxt3 mentioned.

ccrama:
That looks like the best way to do this @PiwwowPants, thank you! 

The OpenContent class is used throughout the app for opening (especially Reddit) links, and just checking if the context is OpenContent will be true for links we also don’t want to act as a new TaskAffinity (like inbox messages). I think I’m going to merge your changes and create a sister class to OpenContent that we can use for only the external intents that we want Slide to handle on top of another Slide instance, which I think should circumvent this issue and still let us use OpenContent internally :)

PiwwowPants:
I'm actually not convinced removing the separate item from recent apps is a good choice. When Slide opens an external app, the context of the new instance is lost. 

FastHub-Libre handles it by just leaving both instances in Recent Apps and that may be a much better way to handle it. Alternatively Twitter just overrides the session like the current behavior. 

I'd probably just recommend either dumping my PR for now or pushing a change to remove the exclusion from recent apps (or wait until I do it)

ccrama:
[...]

PR #2640, opened 2018-01-30
Handle app launched from third party app as separate task

PiwwowPants:
Feature change based on https://github.com/ccrama/Slide/issues/2397
Merge conflict with https://github.com/ccrama/Slide/pull/2639. One of these branches will need a specific change to function together  

When Slide is opened from the user's launcher, the app will behave as it always has.
When Slide is opened via an intent from a third party application, a new task will be created in a separate affinity. This task will always start as a new task and omit itself from recents to avoid having two instances of Slide in the recents app screen.

After closing the task, it cannot be re-opened naturally by the user and history is functionally lost.

Tested:
- Opening Slide from launcher behaves as it did previously
- Opening Slide via Intent from third party app does not replace the standard Slide task
- New Slide affinity returns to previously opened app on back press / swipe back / close
- New Slide affinity is removed from recent apps when closed
- Launching another instance of the new Slide affinity after having already launched one does not retain previous history from old instance

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -33,6 +33,7 @@
         <activity
                 android:name=".Activities.Slide"
                 android:alwaysRetainTaskState="true"
+                android:taskAffinity=".SlideTask"
                 android:configChanges="keyboard|keyboardHidden|orientation|screenSize|smallestScreenSize|screenLayout|uiMode"
                 android:icon="@mipmap/ic_launcher"
                 android:label="@string/app_name"
@@ -50,7 +51,6 @@
                 android:alwaysRetainTaskState="true"
                 android:configChanges="keyboard|keyboardHidden|orientation|screenSize|smallestScreenSize|screenLayout|uiMode"
                 android:label="@string/app_name"
-                android:taskAffinity="me.ccrama.SubsOverview"
                 android:windowSoftInputMode="adjustResize">
         </activity>
         <!--Disable this for now
@@ -70,6 +70,7 @@
  -->
         <activity-alias
                 android:name=".Activities.SlideDefault"
+                android:taskAffinity=".Slide"
                 android:enabled="false"
                 android:icon="@mipmap/ic_launcher"
                 android:label="@string/app_name"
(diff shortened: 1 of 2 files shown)


PiwwowPants:
Not sure if we are going with this branch but I still rebased it to resolve the merge conflict

ccrama:
It appears as if #2632 is causing notifications and messages to open as a new task on top of the current context, and is causing some confusion by having two instances of Slide open at a time from multitasking. If they close the message, doing task switching (two presses on multitasking in Android N+) opens a new version of Slide, presumably because it has the inbox in recents along with another version of Slide in recents, and when the context is deleted after the user leaves the screen (don't know why this is happening, possibly a flag?) Android just starts Slide again with the default intent and now the user has two versions of Slide in recents. Might be a little confusing to visualize, to reproduce just open a message, swipe out, then multitask back to it

ccrama:
Have you been noticing anything like this @PiwwowPants? ^

PiwwowPants:
@ccrama sorry. Didn't see this. I haven't noticed this particular issue but I'm sure it's happening. This implementation isn't very fleshed out and honestly, multiple task handling is a but of a hell scape. 

An issue I have seen that this change brought on some weird slowness when starting up Slide from the home screen. Without this change it opens right away like normal but with it it could take 2-5 seconds just to launch.

ccrama:
@PiwwowPants  definitely is a hellscape... I think the issue may be stemming from FLAG_ACTIVITY_CLEAR_TASK, [this Stackoverflow](https://stackoverflow.com/questions/21833402/difference-between-intent-flag-activity-clear-task-and-intent-flag-activity-task) post does a pretty good job of describing what some of these flags mean in conjunction with each other. I *think* if we just use FLAG_NEW_TASK and also use OpenContent as a separate taskAffinity, we can get the behavior we are looking for

ccrama:
My changes seem to be working correctly, now the only issue is swiping the activity away keeps it alive in recents for some reason. Anyone know what the cause of this could be?

PiwwowPants:
I am unaware of what that issue could be. I was pretty sure the system enforced that behavior pretty aggressively. Do you mean you swipe it out of recents, close it, re-open it, and it's still in recents?

ccrama:
I was actually able to fix it by telling the app not to keep OpenCoontent in recents

PiwwowPants:
To verify, that does not mean that you prevented the task from appearing in recents at all right? Just that now the bug that prevented it from ever leaving is gone?

PR #2643, opened 2018-02-04
Fixed error retry dialogs in Submit to display messages correctly

PiwwowPants:
Not related to any open issue

In the Submit class, error dialogs for "retry" were displaying resource ids since getString was not called on the resource ids for the message creation.

I just condensed all the logic into a single reusable method since all dialogs here do the same thing except minus the error message.

This change went through a few iterations so make sure I didn't screw up any of the messages.


ccrama:
Whoops... Thanks for fixing that! Looks great

PR #2644, opened 2018-02-05
Consolidated sharing logic into a helper class

PiwwowPants:
Fixes [Reddit reported issue](https://www.reddit.com/r/slideforreddit/comments/7v2nc4/v58alpha2_released/dtpetg0/)

Originally, the issue was I didn't replace all the `FileProvider `usages across the app when I created `FileUtil`. In these, the old provider name (.MediaView) was being used. I renamed the provider to be the generic `.provider` name to appease TedBottomPicker

In testing this, I ran into some confusion on why there were multiple `shareImage `methods that seemed to vary only slightly and in confusing ways, with the `MediaView#shareImage` logic just straight up not working. The only `shareImage `method I could get to work was the one in `AlbumPager`. This one was the simplest and had the fewest dependencies. When I converted ALL `shareImage `methods to be this same method, they all worked consistently without an issue.

I do not have answers for what special things those other methods were doing and why they were causing failures in sharing when using the `FileUtil#getFileIntent` method. I'm not ruling something out in how I wrote it, I'm just unsure.

Now, this time I DID test:
- Sharing links
- Sharing images
- Sharing from albums
- Sharing from tumblr albums
- Downloading images (notifications still work)
- Downloading from albums / tumblr albums (notifications still work)
- Downloading gifs (notifications still work)

Please verify that me removing those other methods and consolidating them into a single method does not break any required logic. I am not sure about edge cases they could have been resolving.

Additionally, this `Util `package is getting out of hand. Looking for incorrect usage of static helper classes (my new `ShareUtil `class included) could be a project.

EDIT: Yes I should have broken this into two commits. No it would not have fixed the depressingly large PR.

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Activities/AlbumPager.java
+++ b/app/src/main/java/me/ccrama/redditslide/Activities/AlbumPager.java
@@ -2,9 +2,6 @@ package me.ccrama.redditslide.Activities;
 
 import android.app.Activity;
 import android.app.Dialog;
-import android.app.Notification;
-import android.app.NotificationManager;
-import android.app.PendingIntent;
 import android.content.DialogInterface;
 import android.content.Intent;
 import android.content.res.TypedArray;
@@ -13,16 +10,12 @@ import android.graphics.Color;
 import android.graphics.PorterDuff;
 import android.graphics.Typeface;
 import android.graphics.drawable.Drawable;
-import android.media.MediaScannerConnection;
-import android.net.Uri;
 import android.os.AsyncTask;
 import android.os.Bundle;
 import android.os.Environment;
 import android.support.v4.app.Fragment;
 import android.support.v4.app.FragmentManager;
 import android.support.v4.app.FragmentStatePagerAdapter;
-import android.support.v4.app.NotificationCompat;
-import android.support.v4.content.FileProvider;
 import android.support.v4.view.ViewPager;
 import android.support.v7.widget.Toolbar;
 import android.util.Log;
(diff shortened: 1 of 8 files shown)


app/src/main/java/me/ccrama/redditslide/Activities/AlbumPager.java
```diff
@@ -758,125 +749,6 @@ public void onClick(DialogInterface dialog, int which) {
 
     }
 
-    public void showNotifPhoto(final File localAbsoluteFilePath, final Bitmap loadedImage) {
```
PiwwowPants: I do not really know what this was doing. Removing it seemed to make no change. If you could explain its purpose I can try to test if I broke everything.

app/src/main/java/me/ccrama/redditslide/Activities/MediaView.java
```diff
@@ -1321,133 +1317,6 @@ public void onClick(DialogInterface dialog, int which) {
...
-                doAlternativeImageSave(image, bitmap, original);
-                shareFile(image);
-            }
-        } catch (IOException | NullPointerException e) {
-            e.printStackTrace();
-            Toast.makeText(this, getString(R.string.err_share_image), Toast.LENGTH_LONG).show();
-        }
-    }
-
-    public void shareFile(File image) {
```
PiwwowPants: Dropping this method entirely did not seem to change if sharing succeeded from MediaView. What additional work was this method doing. If something meaningful, should it be re-added so AlbumPager can leverage it too?

ccrama:
All looking good! Thank you for the PR @PiwwowPants  :)

PR #2664, opened 2018-02-25
[Do not merge] Feature/replace bottom sheet with android

PiwwowPants:
DO NOT MERGE. THIS IS EXCLUSIVELY A CODE REVIEW.

I will add to these notes later and ping you when it's fully updated. I am just finished this and pushing RIGHT before I go somewhere but I'd like to get the implementation of the built-in BottomSheetHelper I wrote in your hands now.

The overarching question is, "Does this seem like it can reduce complexity or raise it?". I did not modify (too much) any implementation of the logic on the bottom sheets. I may have added some LongPress actions here and there, but I copy-and-pasted right out of the switch statements.

Ideally, if you want to go with how this is implemented, I would want to go back through, move some common logic somewhere more common, and possible template-ify some of the more common BottomSheet types (e.g., the link sharing one that is used more than once). I also need to put back in the nice "ontouch" tinting. Right now there is no feedback when you press and I have no real clue why yet.

There is no need to read all of this since it won't be merged like this, but I would like your thoughts on the Helper class (uses a pseudo-builder-pattern) and how it is written in the code. A pro tip is to look at a single massive diff (where I ripped out the third part BottomSheet), since they are all like that anyway. I'll do my best to point out oddities. Otherwise just stare at the new files and the resource file changes.


PiwwowPants:
That failed check is lost on me. It's building and running on my side. I will see if maybe something funky happened in that merge from master (that really... I'm stumped on because I rebased...). Maybe I missed something?

ccrama:
Really great stuff, thank you @PiwwowPants ! This is definitely going to take some time to look over, will look into it over the next few days :D

ccrama:
Thank you very much for the PR! After going through the code a bit more carefully, things I like

- A pure Google implementation we can extend for other parts of the UI
- Really great way to build them
- Implementation takes away a lot of boilerplate and it handles themes well
- Love the long-click implementations


Some things I have questions about
- Which should we template for? I can go through and do this if you want!
- Do you have any reservations about this system?
- It has conflicts :D

PiwwowPants:
I think we should template out any common logic. Things like copying links, mod behaviors, etc. This may be something good for you to look at, but I can too. It'll clean up a lot of duplicate code to keep things safer. Whether that means templating an entire sheet or just individual items would be a call by you. 

My reservation is that I used a pseudo-builder pattern. I'm just going to use a real one and handle stuff better in build. I think I really like the pattern since it still allows things to be very structured and you're in full control (even if it becomes a library). The sheets builds in the order the code calls the items too, so you can dynamically control the order of the items.

I fixed the conflicts locally with a Rebase. I can push that up soon. 

You can definitely mess with this and push whatever you see fit. The next hurdle is those templates so we don't have to keep making the same buttons over and over again all over the place. I am confident I made mistakes in ordering and logic too so comparing to the old code will be necessary when doing that.

PiwwowPants:
UPDATE
This definitely is still in progress.
I'm definitely not avoiding looking at the horrible conflicts.

PiwwowPants:
I am closing this until I can get it more worked out. I believe this is just going to have to be a full refactor of all of the bottom sheets all at once to add this feature and consolidate into a new helper. It became large enough that this PR will be irrelevant.

PR #2681, opened 2018-03-09
Android P Support and crash fix

PiwwowPants:
[Reddit issue](https://www.reddit.com/r/slideforreddit/comments/82v9wn/crash_in_android_p/)

This adds a new flavor dimension for Android P.supoort, since you can't install an APK targeting a preview build on a release app. If there's a better way, let me know. You can distribute P builds to beta testers now and it can't be used on other API versions, while the current stable version doesn't target P. 

This also fixes two bugs. The first was in AdBlocker, the domains were throwing an index out of bounds. I can't tell why, but switching the logic to a worker thread reading a text file (in accordance to the link above the logics post), resolved the crash. 

Additionally I fixed a condition in cross posting that felt misplaced. It checked if a value didn't exist on a JSOM string and if it didn't, it tried to get a value from it. This would always throw a null pointer. 

Resolving these bugs let the app run on Android P. I verified the P app runs on P and the stable app runs on P and release versions. Please also verify this personally.


PiwwowPants:
TravisCI will probably fail if you're depending on the build variant names being the same as before. This change changed the build variant names. 

If you want to go this route, you'll want to update Travis to target the apiStable dimension of whatever you want.

app/build.gradle
```diff
@@ -78,65 +88,67 @@ ext {
...
-    compile 'com.makeramen:roundedimageview:2.2.1'
-    compile 'com.getbase:floatingactionbutton:1.10.1'
-    compile 'com.sothree.slidinguppanel:library:3.3.1'
-    compile 'com.github.suckgamony.RapidDecoder:library:7cdfca47fa'
-    compile 'com.github.dasar:shiftcolorpicker:v0.5'
-    compile 'com.squareup.okhttp3:okhttp:3.7.0'
-    compile 'com.google.code.gson:gson:2.8.0'
-    compile 'com.github.ccrama:TedBottomPicker:496623c9b6'
-    compile 'commons-io:commons-io:2.5'
-    compile('com.github.ozodrukh:CircularReveal:1.3.1@aar') { // ↑ major
```
PiwwowPants: When I switched to implementation, I stopped targeting these `aar` files. This let me drop `transitive=true` so I could include it in the array block.

If there was a reason you were targeting `aar` on this and the other let me know.

ccrama: Honestly only reason was because that library specifically required it, but it seems to be working fine without it after a few days of testing!

Issue #2680, opened 2018-03-09
Removing post as a mod causes AppCrash


dylmye:
Hi, 

Removing a post with a reason causes the app to crash. This is as a mod with full permissions. The action is still performed (the post is deleted) but the app force closes ~5 seconds after. 

I've just sent through the ANR report; if there are any more details I can add please let me know. 

OnePlus 3 / OOS Oreo Stock

PiwwowPants:
This only occurs for you when removing a post with a reason, correct? I imagine removing without a reason doesn't cause the crash.

dylmye:
That's correct.

PR #2685, opened 2018-03-11
Pass correct value to API when removing submission with reason

PiwwowPants:
Fix for issue https://github.com/ccrama/Slide/issues/2680

Previously, when removing a post with a  reason, the wrong id value was passed to the method `Authentication.reddit.get(submission.getFullName()).get(0)`, causing an index out of bounds exception from being thrown.

This change passes the correct id, allowing the app to remove a post with a reason without crashing.

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/SubmissionViews/PopulateSubmissionViewHolder.java
+++ b/app/src/main/java/me/ccrama/redditslide/SubmissionViews/PopulateSubmissionViewHolder.java
@@ -58,11 +58,6 @@ import net.dean.jraw.models.VoteDirection;
 
 import org.apache.commons.lang3.StringEscapeUtils;
 
-import java.io.IOException;
-import java.net.HttpURLConnection;
-import java.net.MalformedURLException;
-import java.net.URL;
-import java.net.URLConnection;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


app/src/main/java/me/ccrama/redditslide/SubmissionViews/PopulateSubmissionViewHolder.java
```diff
@@ -1723,11 +1710,11 @@ public void onPostExecute(Boolean b) {
             @Override
             protected Boolean doInBackground(Void... params) {
                 try {
-                    String toGet =
-                            new AccountManager(Authentication.reddit).reply(submission, reason);
+                    new AccountManager(Authentication.reddit).reply(submission, reason);
                     new ModerationManager(Authentication.reddit).remove(submission, false);
                     new ModerationManager(Authentication.reddit).setDistinguishedStatus(
-                            Authentication.reddit.get(toGet).get(0), DistinguishedStatus.MODERATOR);
```
PiwwowPants: Since I run the formatter every time, so fixes get lost in the sea of changes, here is the ONLY functionality change that fixed this.

"toGet" is the id of the submission, but not the name. When Authentication.reddit.get(toGet), runs, it gets the value based on the id. Since this id was not correctly, an index out of bounds of thrown. The post WAS correctly removed prior but the app was crashed.

Passing in `submission.fullName()` resolved this as that is the correct id the method was looking for.

ccrama:
Ah great fix! Thank you @PiwwowPants 

I have returned from vacation, will check out your Android P fix soon :D

PR #2691, opened 2018-03-25
Fix crosspost crash on startup

USA-RedDragon:
The logic flow before was:
  Check if json does not have crosspost_parent_list AND
  get member 0 of json's crosspost_parent_list to check if it's not null

The problem is if the first condition is true(json does not have crosspost_parent_list),
AND can't short-circut and it will move on to evaluate the second side, so it will
try and get member 0 of a crosspost_parent_list, which will be null as we figured out
in the first conditional.

Now the logic flow is:
  Check if json does not have crosspost_parent_list OR
  get member 0 of json's crosspost_parent_list to check if it's null

This way, the OR can short circut on the first true condition, so the get member 0
logic won't trigger unless the json does have crosspost_parent_list.

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/SubmissionCache.java
+++ b/app/src/main/java/me/ccrama/redditslide/SubmissionCache.java
@@ -96,7 +96,7 @@ public class SubmissionCache {
         String spacer = mContext.getString(R.string.submission_properties_seperator);
         SpannableStringBuilder titleString = new SpannableStringBuilder("Crosspost" + spacer);
         JsonNode json = s.getDataNode();
-        if (!json.has("crosspost_parent_list") && json.get("crosspost_parent_list").get(0) != null) { //is not a crosspost
+        if (!json.has("crosspost_parent_list") || json.get("crosspost_parent_list").get(0) == null) { //is not a crosspost
             return new SpannableStringBuilder();
         }
         json = json.get("crosspost_parent_list").get(0);


PiwwowPants:
This same fix exists in my Android P pull request https://github.com/ccrama/Slide/pull/2681

For a compare and contrast in case we did anything different.

ccrama:
Updating Google Play has caused issues in the past, have you done any testing with Drive for this PR? Also I believe running gradle in parallel breaks our CI, any reason you enabled that?

Thank you for the PR @USA-RedDragon !

USA-RedDragon:
I didn't mean to merge the second commit. I'll go and reset HEAD and remake the pull request since Github is weird like that

PiwwowPants:
@USA-RedDragon git reset --hard HEAD^ | git push -f origin

Fast forwards. The way to live.

USA-RedDragon:
Gerrit is my poison of choice, so GitHub seems archaic now. Fast forwards are evil lol

USA-RedDragon:
There we go. Removed the personal commit

ccrama:
Thank you for the fix, will merge this now :)

Alexendoo:
I don't think it would break CI but if it does I can take a look at it, what was the change exactly @ccrama?

ccrama:
@Alexendoo  the change was removed so no worries! We will have to update travis for @PiwwowPants #2681 though, probably going to merge that today or tomorrow as I've been testing for a week with no real issues yet

Issue #2690, opened 2018-03-25
Crash when SubmissionsView refreshes with the latest adapter content out of view


PiwwowPants:
[Related Reddit Issue](https://www.reddit.com/r/slideforreddit/comments/86tlho/app_crash_when_refresh/)
[Potential Resolution from StackOverflow](https://stackoverflow.com/questions/31759171/recyclerview-and-java-lang-indexoutofboundsexception-inconsistency-detected-in)

When on a subreddit that contains multiple pages of submissions (e.g. /r/all), a user can scroll and the RecyclerView accepts new data as it comes in. When the user reaches the end of this content, they see "No more Content. Refresh". **Pressing "Refresh" then scrolling up quickly before refresh finishes causes Slide to crash.**

Relavant LogCat line:
`java.lang.IndexOutOfBoundsException: Inconsistency detected. Invalid item position 533(offset:533).state:607 android.support.v7.widget.RecyclerView{1278297 VFED.V... ........ 0,0-1426,2614 #7f10010a app:id/vertical_content}, adapter:me.ccrama.redditslide.Adapters.SubmissionAdapter@c289e73, layout:me.ccrama.redditslide.Views.CatchStaggeredGridLayoutManager@f66cc84, context:android.support.v7.view.ContextThemeWrapper@a3f085f`

This _seems_ to happen from the SubmissionsView. I am looking into it now but wanted to post this up in case I fail.

Let me know if you need anything else to work off of.

PiwwowPants:
Seems to occur in the LoadData AsyncTask. Line 445 of SubredditPosts
```
posts.addAll(filteredSubmissions);
posts = new ArrayList<>(new LinkedHashSet(posts));
offline = false;
```

Attempting to scroll after this logic runs immediately causes the crash.

PiwwowPants:
My thought here is to "resolve", we just call `SubmissionsView.forceRefresh()` in some way. Would require making the SubmissionFooterViewHolder aware of the view it is in though.

PiwwowPants:
Final issue. When on the last view, the ActionBar could not be displayed so when scrolling to the top, it is not shown again. I noticed the ToolBar in the MainActivity does not have the `app:layout_scrollFlags="scroll|enterAlways"` flags that handle scrolling it out of view. Instead a custom handler is used.

Should the custom handler continue to be used or replaced for this flag?

ccrama:
The custom handler is required for some API versions and for the CommentPage view type @PiwwowPants

PR #2692, opened 2018-03-25
Bugfix/submissionview refresh recyclerview crash

PiwwowPants:
Resolves issue https://github.com/ccrama/Slide/issues/2690

I decided to keep the ToolbarScrollHideHandler in this change instead of trying to replace it. It works and there's no reason to remove it. I just worked around it a bit.

The issue explains the problem and resolution pretty well but generally:

- If the content in the adapter for the RecyclerView is being updated on another thread and a user tries to interact with the RecyclerView (scrolling up out of range of the latest content in the adapter), an IndexOutOfBounds would occur
     - To reproduce, reach the end of a subreddit's post listing, press "Refresh" and scroll up
- This change does two things to resolve
    - It consolidates the logic for refreshing the view by calling the method SubmissionView.forceRefresh() from the SubmissionAdapter. I explicitly cast the type because I can't fathom why this adapter would be used elsewhere. There are better patterns for this and if you can think of one easily added, let me know
    - The List (?) storng all the Submission data is cleared on forceRefresh, the view is pushed to the top, and the toolbar is forcibly shown. This prevents the user from scrolling into soon-to-be stale data on refresh
        - In this, I added a new method to the SubmissionDisplayer interface to make it easier. All implementations now have a method that lets the View know that the adapter was changed in someway from somewhere else. This is helpful when the changes happen outside of the adapter, but the view is still available (like a hacky MVP).

I separated the commits in a functionality change and a cleanup change to make it easier to see what had to change

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Fragments/SubmissionsView.java
+++ b/app/src/main/java/me/ccrama/redditslide/Fragments/SubmissionsView.java
@@ -443,7 +443,8 @@ public class SubmissionsView extends Fragment implements SubmissionDisplay {
     }
 
     public void forceRefresh() {
-        rv.scrollToPosition(0);
+        toolbarScroll.toolbarShow();
+        rv.getLayoutManager().scrollToPosition(0);
         mSwipeRefreshLayout.post(new Runnable() {
             @Override
             public void run() {
@@ -540,6 +541,11 @@ public class SubmissionsView extends Fragment implements SubmissionDisplay {
                 }
             }
         }
+    }
+
+    @Override
+    public void onAdapterUpdated() {
+        adapter.notifyDataSetChanged();
     }
 
     public void resetScroll() {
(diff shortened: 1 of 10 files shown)


app/src/main/java/me/ccrama/redditslide/Fragments/SubmissionsView.java
```diff
@@ -443,7 +443,8 @@ private void refresh() {
     }
 
     public void forceRefresh() {
-        rv.scrollToPosition(0);
+        toolbarScroll.toolbarShow();
```
PiwwowPants: This was necessary because `scrollToPosition` calls a LayoutChange even, not a Scroll event, so we couldn't directly leverage the existing logic on the Helper.

ccrama:
> In this, I added a new method to the SubmissionDisplayer interface to make it easier. All implementations now have a method that lets the View know that the adapter was changed in someway from somewhere else. This is helpful when the changes happen outside of the adapter, but the view is still available (like a hacky MVP).

Great addition as well

Issue #2699, opened 2018-03-30
Crash when opening post


incIined:
Slide v5.8.5-a1
Android 8.0 on TouchWiz, 2/1/18 security patch level

Post in question: https://redd.it/88749u

Video of what happens: https://streamable.com/zjbqi

No logs, I'm not rooted and I don't have access to a computer atm.

incIined:
Fixed! Thanks y'all.

PR #2700, opened 2018-03-30
Bugfix/crash on null flair text

PiwwowPants:
Resolves issue https://github.com/ccrama/Slide/issues/2699
Additionally it resolves [this reddit issue](https://www.reddit.com/r/slideforreddit/comments/885mil/v585alpha1_released_firefox_support_flair_fix_and/dwi61zu/) which I believe is the same issue with flairs.

Previously, the logic for checking if flair was null did this check

```
if (comment.getAuthorFlair() != null 
    && !comment.getAuthorFlair().getText().isEmpty() 
    && !comment.getAuthorFlair().getText().isEmpty())
```

This would allow a null pointer to get thrown since you'd be checking isEmpty on a potentially null value. I modified the check to see if getText is null before checking isEmpty as well as consolidated the code a bit.

As always, preview the two commits separately to see the functionality change without the cleanup in the way

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Adapters/CommentAdapterHelper.java
+++ b/app/src/main/java/me/ccrama/redditslide/Adapters/CommentAdapterHelper.java
@@ -336,7 +336,7 @@ public class CommentAdapterHelper {
                                     Snackbar.LENGTH_SHORT);
                         }
                         View view = s.getView();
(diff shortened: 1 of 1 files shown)


ccrama:
I thought I had re-added the null check, but must have missed a few! Thank you for the quick work :)

PR #2702, opened 2018-03-31
Added null checks to ensure json values aren't null before utilizing

PiwwowPants:
[Fixes Reddit issue](https://www.reddit.com/r/slideforreddit/comments/88h1qr/rwholesomegreentext_crashes_slide/)

When starting a post, it's possible (for some reason), that a crosspost_parent_list node is not null but contains a null value. This change adds a null check on `get("crosspost_parent_list).get(0)` in SubmissionCache and PopulateSubmissionViewHolder to ensure we aren't doing operations on a null value. 

I don't know why Reddit would want to return `"crosspost_parent_list": []`, but it sure does

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Adapters/CommentAdapter.java
+++ b/app/src/main/java/me/ccrama/redditslide/Adapters/CommentAdapter.java
@@ -38,7 +38,6 @@ import android.widget.TextView;
 
 import com.afollestad.materialdialogs.AlertDialogWrapper;
 import com.devspark.robototextview.RobotoTypefaces;
-import com.devspark.robototextview.RobotoTypefaces;
 import com.lusfold.androidkeyvaluestore.KVStore;
 import com.mikepenz.itemanimators.AlphaInAnimator;
 import com.mikepenz.itemanimators.SlideRightAlphaAnimator;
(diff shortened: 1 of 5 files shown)


ccrama:
Thanks for the quick fix! I wonder why that started ocurring, thought my changes earlier this week would have accounted for that 🙄

And my Mac has already officially shit itself 3 months into life so will hopefully get this out with a new update on Monday

PiwwowPants:
I muddied the pull request with another fix
Apparently that flair check was happening elsewhere and failing for the same reason. There may be value in handling Json responses in a more universal fashion to prevent this, but for now, I believe I fixed them all. 

This is likely the cause of the additional crashes people are seeing. Should be good now.

ccrama:
It might be worth looking into implementing the new JRAW library with Kotlin which does handle these cases a little more gracefully. But thank you for the continued fixes, my computer is working again so will push this out soon :D

PiwwowPants:
I literally just remembered the Guava libraries have `Strings.isNullOrEmpty()` today. We most definitely should just be using that more appropriately.
I almost did that today then saw all the usages of `.isEmpty()`. May be good to just remember going forward.

Nxt3:
Has anyone considered using Kotlin in some areas? Since it's fully compatibkr with Java, you could rewrite a bit of Slides code and prevent all these `if (!null)` checks.

ccrama:
@Nxt3  if we switched to the new Kotlin version of JRAW and went with a MVP architecture using Realm my life would be complete

PR #2714, opened 2018-04-07
[Bugfix] Check for null before isEmpty()

jseybold:
Affects branch/tag 5.8.5 and master

```
java.lang.NullPointerException: Attempt to invoke virtual method 'boolean java.lang.String.isEmpty()' on a null object reference
       at me.ccrama.redditslide.Fragments.MediaFragment.onCreateView(MediaFragment.java:172)
       [...]
   Force finishing activity me.ccrama.redditslide/.Activities.Shadowbox
   Force finishing activity me.ccrama.redditslide/.Activities.MainActivity
 Showing crash dialog for package me.ccrama.redditslide u0
```


app/src/main/java/me/ccrama/redditslide/Fragments/MediaFragment.java
```diff
@@ -169,11 +169,12 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container,
                 ((ImageView) rootView.findViewById(R.id.thumbimage2)).setImageResource(
                         R.drawable.nsfw);
             } else {
-                if (firstUrl.isEmpty() && !s.getThumbnail().isEmpty()) {
-                    ((Reddit) getContext().getApplicationContext()).getImageLoader()
-                            .displayImage(s.getThumbnail(),
-                                    ((ImageView) rootView.findViewById(R.id.thumbimage2)));
-
+                if (firstUrl == null || firstUrl.isEmpty()) {
+                    if (s.getThumbnail() != null && !s.getThumbnail().isEmpty()) {
```
PiwwowPants: Probably return this into a single line with an `&&`.
Also a note. We don't seem to use it anywhere, for better or worse, but the common libraries have `Strings.isNullOrEmpty()` that would consolidate these checks down.

jseybold: I split it for readability since the first check is for null/empty and the second check is for **not** null/empty. If we have access to those library calls here, we should absolutely be doing that. This is like the third bug in the last week that I've found related to null/empty checking.

I'll update this PR soon.

PiwwowPants: Gotta make sure those alpha testers are actually opening the app
Keeping them on their toes.

jseybold: Actually I've been stumbling into them in the release builds. Most of them have already been fixed in the alpha's, except this one.

>.<

ccrama: @jseybold unfortunately some of those bugs were caused by API changes on Reddit's side and then Reddit undoing those changes between builds. Thankfully though this is one of the only NPEs currently reported by users on v5.8.5 (other ones are NPEs in internal library code...)!

jseybold:
Updated with Strings.isNullOrEmpty()

PR #2719, opened 2018-04-10
Added settings option to open the navigation drawer on back press

PiwwowPants:
This add an option like in Discord where pressing back from the MainActivity will not close the app and instead will open the navigation drawer.

I do not know why cleanup decided to remove all those indentations (but not ALL of them) but it did. I can remove that piece of the change if wanted. I promise I didn't delete any  

Yes it works!


PiwwowPants:
Agreed. I can look into the UX for that when I get a chance. Thanks!

PiwwowPants:
I am not impressed with myself in that I did not merge down any of the duplicate logic, but this behaves now like the "Subreddit search method" option on the same screen.

Just to get it up and going this will work well, in my opinion.

ccrama:
I would like the default to be prompt before close if possible, looks great though! Thanks!

PiwwowPants:
Sounds good. I made that change from GitHub's code editor on my phone!
That means it's good.

ccrama:
@PiwwowPants does that also set the default value in SettingValues.java? :D

ccrama:
I was tripping, it's all good!

PR #2723, opened 2018-04-17
Confirm with user before backing up / restoring from google drive

YusefOuda:
Recently, I accidentally clicked "backup" on a new install of Slide so all my previous backed up settings were gone. This PR adds a dialog to have users confirm whether they want to backup/restore.

Let me know if you have any feedback/concerns/etc.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -1,5 +1,6 @@
 <resources>
     <!-- Generic text used anywhere -->
+    <string name="general_confirm">Confirm</string>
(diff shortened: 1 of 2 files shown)


ccrama:
Great addition, I am also going to implement backup "versions" soon. Thank you!

YusefOuda:
Cool, that will be great.

Just noticed a typo in both strings, they say "will over overwritten" instead of "will be overwritten", do you mind adding in that quick fix? Otherwise I can do so late tonight with another PR

PR #2726, opened 2018-04-18
Add discord badge

Alexendoo:


Alexendoo:
I changed the link in the sidebar of /r/slideforreddit, can be done for the iOS sub too. May or may not want to make announcement posts

ccrama:
Thank you @Alexendoo ! I will probably make a new channel for the iOS version to keep the chat cleaner

PR #2727, opened 2018-04-21
Fix two remaining htts -> https

timawesomeness:
https://www.reddit.com/r/slideforreddit/comments/8du9r2/share_comments_uses_htts_instead_of_https_in_a/


ccrama:
Thanks for the fix! I should have made this a method instead 😅

PR #2728, opened 2018-04-21
Guarantee opening an external browser from the internal browser

PiwwowPants:
Partial fix on issue [from Reddit](https://www.reddit.com/r/slideforreddit/comments/8cn2ck/slide_locking_up_my_phone/)

When pressing the "open in external browser" button from the internal browser, the url is opened in the default link handler. On links like "https://www.reddit.com/gold/about", this will cause it to infinitely open back in the Slide app, since that's the default link handler, but that url can only be opened in a WebView.

In internal browsers, this is manageable. When Slide is set to open in an external browser, this would likely cause an infinite loop, locking the phone.

This fix only fixes the button in the internal browser now to get your opinion. I added an intent on a URL that will likely never be link handled. This guarantees the returned package is a browser, not another app (or Slide itself). This may not be desired behavior. I could convert it so if the package to be opened is Slide, we open in an external browser, otherwise, let it be link handled.
Whatever we decide, I can try to convert to the other ways Slide opens urls, to fix the infinite loop

Hopefully any of that makes sense.

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Activities/Website.java
+++ b/app/src/main/java/me/ccrama/redditslide/Activities/Website.java
@@ -33,9 +33,9 @@ import me.ccrama.redditslide.PostMatch;
 import me.ccrama.redditslide.R;
 import me.ccrama.redditslide.Reddit;
 import me.ccrama.redditslide.SettingValues;
-import me.ccrama.redditslide.Views.NestedWebView;
 import me.ccrama.redditslide.Visuals.Palette;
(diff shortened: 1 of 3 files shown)


PiwwowPants:
I actually made the change to only force into an actual browser if the package used for the URL is Slide. Otherwise, it will use default link handling behavior.

Additionally I added access to static application context (without leaking Context!) so helper methods do not require Context passed in when they realistically just need application context.

ccrama:
Great work, thanks for the PR!

PR #2730, opened 2018-04-21
Attempt to prevent Slide from opening a URL in itself

PiwwowPants:
Lemme try this again

This change does three things:
- Prevents Slide from opening itself when using external link handling
- Uses static application context in the LinkUtil.openExternally method
- Uses LinkUtil.openExternally when opening externally from an internal browser

If you could test something for me:
1. Set Slide to open links Internally
2. Open the Reddit link from [my comment here](https://www.reddit.com/r/slideforreddit/comments/8cn2ck/slide_locking_up_my_phone/dxqe0rn/) (from within Slide)
3. Post result


PiwwowPants:
Still a meaningful change regardless of https://github.com/ccrama/Slide/pull/2740

I'd just have to resolve the merge conflicts on either branch that merges second.

PR #2732, opened 2018-04-22
Feature/select non default browser

PiwwowPants:
This change actually piggy-backs off of https://github.com/ccrama/Slide/pull/2730. Because of this, that one should probably merge first (I literally merged it into this branch).

This change removes the Firefox option for link handling. Instead, it adds a picker to the "External browser" option that lets you choose any external browser. This is only available on API 23+. Per Alex, 23+ covers 92% of all Slide users.

If a user is API 22- or only has on browser OR never selected a browser for this mode, it will default to the default browser.


PiwwowPants:
I tried to call out the bits that matter, since, until that other PR merges, this one is a bit unmanageable to read. It'll get easier when / if that goes in.

PR #2733, opened 2018-04-22
Searchable Settings Menu

jseybold:
As the title suggests, this PR contains changes to the settings menu(s) to make them searchable. Not every sub menu will be searchable since some of the individual settings don't make much sense without context. The general design is to add a 'Search' EditView to the main settings page and dynamically update the view beneath it depending on the text that is entered.

Settings sub menus are having most of their functionality moved to a Fragment that accepts an Activity in its constructor. This allows the Fragment to be used by the sub menu Activity as well as by the Settings Activity.

Layout xml files have also been split into a parent/child model. The android:id tags have also been deconflicted since this feature makes it possible to combine two different layouts that contain an element with the same id. Text that is not meant to be clicked has been given an android:tag="label" attribute to prevent their parent ViewGroup from being removed. Without this, the pruning is overly aggressive and will remove matches that should be displayed to the user in the search results.


app/src/main/java/me/ccrama/redditslide/Activities/Settings.java
```diff
@@ -33,33 +53,128 @@
...
+        BuildLayout(prev_text);
+    }
+
+    @Override
+    protected void onSaveInstanceState(Bundle outState) {
+        outState.putInt("position", scrollY);
+        outState.putString("prev_text", prev_text);
+        super.onSaveInstanceState(outState);
+    }
+
```
PiwwowPants: Think it would be worth it here now or in the future to just loop over all the available layouts we want to deal with and run addView for each item? It'd just be a static array of Integers anyway.

app/src/main/java/me/ccrama/redditslide/Activities/Settings.java
```diff
@@ -132,55 +304,73 @@ public void onClick(DialogInterface dialog,
             });
         }
 
-        findViewById(R.id.general).setOnClickListener(new OnSingleClickListener() {
+        ((EditText) findViewById(R.id.settings_search)).addTextChangedListener(new TextWatcher() {
+
+            @Override
+            public void beforeTextChanged(CharSequence charSequence, int i, int i1, int i2) { }
+
```
PiwwowPants: I'm sure there is a real reason but listeners in Android are hell states because the lifecycle can just call these for haphazard reasons.

PiwwowPants:
Please merge this before merging https://github.com/ccrama/Slide/pull/2740 to prevent merge conflict hell that may appear on this PR. It'd be easier for me to resolve mine.

app/src/main/java/me/ccrama/redditslide/Activities/Settings.java
jseybold: You mean once during initialization? Currently I remove basically everything (that may already have been pruned down), re-add it, then prune it again every time the search text input changes.

PiwwowPants: I meant more for code simplification / clarity than anything. This method just called addView for every view we want added so it chains and compounds as more things are potentially added.

jseybold: Oh I see, just do it iteratively so that all that would change are the array contents.

app/src/main/java/me/ccrama/redditslide/Activities/Settings.java
ccrama: Honestly this might have just been from a code template, unless it disqualifies the class as the listener I think we should remove it

jseybold: TextWatcher is an interface, so I believe all of the methods have to be implemented -- even if they don't do anything.

ccrama: Awesome, what I thought

app/src/main/java/me/ccrama/redditslide/Activities/Settings.java
jseybold: Done

ccrama:
For additional cleanup we should remove the current Setting activities that are still in the Activities subfolder

ccrama:
Just merged it with a few UI changes! Thank you very much for this pr :)

PR #2740, opened 2018-04-28
Add option to select any installed browser

PiwwowPants:
Removed Firefox specific option
Converted browser type option to radio group
Moved browser selection into pop up menu
Only force setPackage on intents when required

Here's the change. This will let the user select any browser on their device to use to open links in Slide. This works on any API version.

If a user only has 1 browse, they will not see the option and Slide will continue how it normally does. If they have two, the default browser will be used as it does now until they set one discretely. If they delete their browser they set, it is handled in the LinkUtil.overridePackage method and will fall back to regular behavior.

Additionally, I made it so Intent.setPackage is only run when the package NEEDS overridden and not for every single intent. It only NEEDS it when Slide is the package to be opened or when overriding the browser. Otherwise, it will fallback to default behavior.

The layout for setting a browser type has changed. I moved it out to a RadioGroup and made the actual WebBrowser app selection a pop up. This is a hope of giving more context that these two items are linked.

Let me know if you want anything extra explained


app/src/main/java/me/ccrama/redditslide/Activities/SettingsHandling.java
```diff
@@ -36,11 +38,6 @@
     @Override
     public void onCheckedChanged(CompoundButton buttonView, boolean isChecked) {
         switch (buttonView.getId()) {
-            case R.id.web:
```
PiwwowPants: This case confused me what it was doing since the rest of the browser_type menu options were not in here. Can you explain it? Removing it seemed to not affect anything.

app/src/main/java/me/ccrama/redditslide/util/LinkUtil.java
```diff
@@ -282,19 +243,37 @@ public static void crosspost(Submission submission, Activity mContext) {
         mContext.startActivity(new Intent(mContext, Crosspost.class));
     }
 
-    public static String getPackage(Intent intent) {
+    public static void overridePackage(Intent intent) {
```
PiwwowPants: The flow on this is now

1. Get package name for the given intent
2. Get package name for a browser
3. If the intent package name is equal to Slide's package name, set the packageToSet on the intent to the browser package name
4. If the browser package name is being used, and the user has the browser set up to be overridden, override that package with the user selected one if it still exists on the device
5. If the package is to be overridden, override it.

ccrama:
Amazing work, thank you!

PR #2742, opened 2018-04-30
Added basic picture in picture support for FullscreenVideo

PiwwowPants:
This is hyper-basic picture in picture support for the FullscreenVideo activity. It only works for Android N+ (I believe doing it before N was a pain and not really ideal).

Notes:
1. I put an option to enable it on N+ but noted it as experimental
2. There are no controls on the thumbnail. We are dealing with a WebView so I let the API handle everything until that can be sorted / if we care enough to keep this feature
3. There is a bug that causes the FullscreenVideo activity to run onStop() twice, throwing a non-breaking exception. I do not know why but am looking into it. It does not seem to hurt the app
4. Running Slide with PiP on is SLOW. There's a lot of data rendering all at once. I'll keep experimenting to see what can be improved if anything. If it can't maybe we can dump it.

Notes above are why I kept this minimal. I do not know how meaningful this feature is, who would use it instead of just passing off to YouTube, and if it's worth the pain points for implementation (ideally at some point we'd want to deal with the video better than a straight WebView)

Let me know what you think. "I don't like it" is acceptable since it was not much effort to write any of this.


PiwwowPants:
This totally screws up split screen mode. Maybe this will stay open while I find bugs. Oops

PiwwowPants:
Found a cool bug where the back button doesn't work sometimes. Don't know why because it's only sometimes. I'm sure I broke that logic :disappointed:

ccrama:
Should be noted this only works with the YouTube player if you don't have the plugin installed

PR #2746, opened 2018-05-05
Added all dark themes to Night Mode options

PiwwowPants:
Refactored theme selection simplify logic and force consistency
[Request from Reddit](https://www.reddit.com/r/slideforreddit/comments/8gbpo6/i_would_love_to_get_the_deep_base_theme_as_a/)

I added all non-light color theme options to the "Night Mode" menu. This includes adding the new "Deep" theme as an option.

Additionally I tried my best to clean up a lot of the logic. I added an enum to `ColorPreferences` to hold all the int values per base-theme type. These values are used in both the Theme enum and the new themePairList list used to link theme `resIds` to themeType `ints`

Entire blocks have been condensed into single for-loops around this new themePairList. The logic behaves the exact same and I seem to not be missing anything crucial. Testing it out makes it seem fine (Did not test whatever happens in Tutorial?)

Let me know what you think.

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/ColorPreferences.java
+++ b/app/src/main/java/me/ccrama/redditslide/ColorPreferences.java
@@ -4,9 +4,13 @@ import android.content.Context;
 import android.content.SharedPreferences;
 import android.graphics.Color;
 import android.support.v4.content.ContextCompat;
+import android.util.Pair;
 
 import org.apache.commons.lang3.StringEscapeUtils;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
 import java.util.Locale;
 
 import me.ccrama.redditslide.Activities.Slide;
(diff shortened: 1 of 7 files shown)


app/src/main/java/me/ccrama/redditslide/ColorPreferences.java
```diff
@@ -602,130 +606,207 @@ public int getColor(String s) {
     }
 
     public enum Theme {
```
PiwwowPants: This block looks scary but all I did was swap the hardcoded themeType ints with the new enum values for consistency

app/src/main/res/layout/nightmode.xml
```diff
@@ -67,19 +75,19 @@
                             android:textColor="?attr/fontColor"/>
 
                     <RadioButton
-                            android:id="@+id/blue"
```
PiwwowPants: On these changes, I just moved all the options in Night Mode around to match their order in the Base Theme menu. All that were there remain. Just added Pixel and Deep as options

PiwwowPants:
Both AMOLED options in the settings are broken for reasons unknown to me. Maybe they're not correctly added to the list of items. I'll check on that.

ccrama:
Much needed, thanks!

PR #2747, opened 2018-05-06
Added top-padding to offline mode content view

PiwwowPants:
Adding padding around card view in post layout view
Corrected some xml linter issues

The main changes here were:
1. I added 16dp padding to the top of the "Manage offline content" view. As it was, the text was bumping really close to the navbar
2. I added 4dp padding around the entire card view in the "Post layout" view. It was pushed against the nav bar which was both jarring and sub-realistic (in the main layouts, there is padding around the card)

What came from this additionally:
1. I resolved any linter issues that were not API conflicts, including: 
    - Removing redundant Layouts (lots of LinearLayouts nested in RelativeLayouts)
    - Settings padding to surround entire views instead of specific sides
    - Settings marginRight on views that only contained marginEnd for compatibility reasons
2. Resolved redundant usages of padding in subviews of layouts as a result of changes in the above
3. Removed a hardcoded height values off of linear_layouts in "Manage offline content" view to use wrap_content instead

To my eye, the layouts look identical to how they were, just with my added padding stated above in the "main changes" section

I can get before-after pictures if you need them / this is too tedious

### DIFF
--- a/app/src/main/res/layout/activity_manage_history_child.xml
+++ b/app/src/main/res/layout/activity_manage_history_child.xml
@@ -5,255 +5,207 @@
     android:layout_width="match_parent"
     android:layout_height="match_parent">
 
-    <RelativeLayout
-        android:id="@+id/manage_history_autocache"
-        android:layout_width="match_parent"
-        android:layout_height="56dp"
-        android:background="?android:selectableItemBackground"
-        android:gravity="center_vertical"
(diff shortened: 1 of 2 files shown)


PiwwowPants:
I promise that was the last force-push...

ccrama:
Much cleaner way to do it, thank you for the PR!

PR #2748, opened 2018-05-09
adds button to bottom nav bar to expand/collapse comments

YusefOuda:
This PR adds another icon to the bottom nav bar which when clicked, will either collapse or expand all comments.

I'm not 100% sure if the bottom nav bar is the best place for this, but I couldn't think of where else it could go.

This feature might not be a good idea at all, just thought I would throw it out there to see if it makes sense


ccrama:
Thank you for the PR! What would be your use case for this button vs the button that exists in the 3 dot menu already?

ccrama:
Ah gotcha. I wonder if it would be worth making this either a settings option, or enabled by default when you have collapsed by default set. What are your thoughts on this?

YusefOuda:
A setting seems like a good way to incorporate this

ccrama:
@YusefOuda there are now some conflicts due to the new settings screen, if you would like me to take a stab at it let me know!

YusefOuda:
Sure thing, if you have time to take a look that would be great. I won't have time for a few days here.

ccrama:
Got it implemented, thank you very much for the PR!

PR #2749, opened 2018-05-10
Bugfix/remove empty parameters from urls before using youtube plugin

PiwwowPants:
Attempting to open the links in this Reddit post causes the Slide YouTube plugin to fail to load the URL:
https://www.reddit.com/r/halo/comments/8i8qbt/theres_this_particular_sound_from_halo_3s/

This is because the parameter `&t=` is empty. This caused me to wrote a method to remove any empty parameters from a url before trying to give it to the video plugin (this is the only place I use this method since most browsers probably do not need this).

Additionally, I:
* Moved the YouTube Slide package and class strings out to the donottranslate file to prevent mistypes (verified they still work!)
* Moved the attempt to open the Slide video plugin out to a helper method that returns a boolean if it (hopefully) succeeded.
* Changed the logic after the video plugin does NOT open to respect the user's settings for browsers to open a video (previously we only opened externally forcibly)
    * This change means anyone using Reader mode will be pushed into a view telling them to enable JavaScript (on YouTube at least) since Reader mode won't support this. I do not know how to handle this naturally (**PLEASE WEIGH IN ON THIS COULD BREAK USERS**)


app/src/main/java/me/ccrama/redditslide/util/LinkUtil.java
```diff
@@ -167,13 +171,30 @@ public static Uri formatURL(String url) {
         }
 
         Uri uri = Uri.parse(url);
-        Uri toReturn;
-        try {
-            toReturn = uri.normalizeScheme();
-        } catch (NoSuchMethodError e) {
-            toReturn = uri;
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN) {
```
PiwwowPants: I forgot to mention I did this too. Same behavior but does the check without having to try-catch

xnaas:
Remember, this is all Aegisfate#7418's fault.

ccrama:
I think this might be better for me to include in the plugin itself, although I already do a check for this with that code, hmm.

PiwwowPants:
@ccrama I agree it's reverse fixing the plugin a bit but having a way to clean a URL before use isn't the worst. Most applications can resolve this but the method could be extended to pull certain information out of a URL as well for parsing.

ccrama:
I really like the extracting of the url strings as well :D 

Seems to be a few merge conflicts but I will take a look at them!

PiwwowPants:
Yeah. In the future we can extend it for use across the app since there are a few places that use this sort of logic. Would be nice to consolidate since URLs are (hopefully) uniform by name

Issue #2737, opened 2018-04-25
Suggestion: different themes based on different accounts


incIined:
Like how Telegram X handles themes: if you log into one account, it'll display one theme. Switch accounts, it'll change to another theme. Would be useful for differentiating which account you're signed into on the fly without having to look for your account name.

ccrama:
Done ^!

PR #2750, opened 2018-05-11
Set app theme per-user account

PiwwowPants:
Request from here https://github.com/ccrama/Slide/issues/2737

This change adds per-user account app theme-ing without breaking (hopefully) current themes for everyone.
The theme is now set per-user account and retrieved per-user (unless a user theme does not exist, then it falls back).

This means:
* User themes will (hopefully) remain as they are now
* Any time a user sets a theme in the future, it will set only for that user account, leaving all other accounts unchanged (includes Guest?)

I did not extend this to fonts, since those seem more granular and not as beneficial to set per-user account.


ccrama:
Thank you for the pr! Looks like a good way to implement it

Issue #2731, opened 2018-04-22
Suggestion: add comment account switcher to FAB reply pop-up


incIined:
When replying from the save/vote bar:
![](https://i.imgur.com/86bkRs1.jpg)

when replying from the FAB:
![](https://i.imgur.com/KbdOsTm.jpg) - notice you can't switch account from here. 

Would be pretty useful and more consistent imo. 

Slide v5.8.6

ccrama:
Done by @PiwwowPants !

PR #2751, opened 2018-05-11
Allow switching account from fab reply button

PiwwowPants:
Remove unnecessary ReplyTaskComment from CommentPage
Move "Switch to another account" string to resources

TIL You can switch accounts while replying. Suggestion from here: https://github.com/ccrama/Slide/issues/2731

The reply fab now uses the normal menu for commenting instead of the "Edit" menu which gives it access to account switching while using the fab. I did have to explicitly hide the menu options though.

Additionally: 
* I just use the `ReplyTaskComment` from `CommentAdapter` and deleted the one in `CommentPage` since it is now unnecessary.
* I hide the reply fab if the post is not reply-able

Looks like I broke nothing.


app/src/main/java/me/ccrama/redditslide/Fragments/CommentPage.java
```diff
@@ -316,7 +314,7 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container,
         toolbar = v.findViewById(R.id.toolbar);
         toolbar.setPopupTheme(new ColorPreferences(getActivity()).getFontStyle().getBaseId());
 
-        if (!SettingValues.fabComments) {
+        if (!SettingValues.fabComments || archived || np || locked) {
```
PiwwowPants: This is where I am trying to prevent the reply FAB from showing for posts where you cannot reply. I only vaguely know what I did here, but it seems right.

ccrama: This would be correct, as the archived/np/locked are set onCreate, and this is in onCreateView!

PR #2756, opened 2018-05-13
Moved gradle dependencies out of arrays

PiwwowPants:
All this change does is moves all of the gradle dependencies out of the arrays.

Originally I added them into arrays, because that made sense. But when inside arrays, gradle won't warn you that there are newer versions of a dependency. It seems like a fault in gradle, but since it's not hard to just not have arrays, I dumped the idea.

Everything else is the same (I DID add `//noinspection GradleDependency` above the `material-dialogs` dependency

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -78,61 +78,56 @@ ext {
     robolectricVersion = '3.3.1'
 }
 dependencies {
-    withGPlayImplementation(
-            'com.google.android.gms:play-services-drive:10.2.0',
-    )
+    withGPlayImplementation 'com.google.android.gms:play-services-drive:10.2.0'
 
-    implementation(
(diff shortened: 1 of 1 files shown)


ccrama:
Is there any particular reason for this? I personally think having it in an array looks cleaner

PiwwowPants:
I already responded on Discord but to keep a single source of truth:

"If you put them in arrays you won't get linter warnings that you're on an out of date version. Each dependency has to be separate for that
I think the warnings out weigh the cleanliness. But it's up to you since I don't update those"

To Alex's point, maybe it would be worth opening an issue with JetBrains about the linter and arrays in Gradle files

ccrama:
Perfect, thanks for the explanation!

Issue #2764, opened 2018-05-14
Hide items in side menu


azumukupoe:
For example, I don't use multireddits so I don't need it in the menu.

Alexendoo:
Closed by #2767

PR #2767, opened 2018-05-15
Added options to hide drawer items

PiwwowPants:
Request from here https://github.com/ccrama/Slide/issues/2764

This change adds a setting to General Settings that shows a dialog to let you turn buttons in the slide out menu on and off. These values are stored as a bitmasked (byte masked? Bite Masqued) long in settings we can read out of from a single source.

This will let you hide a subset of the options (You, Inbox, Multireddits, Go to profile, Discover). All of these are handled in a single `for-loop` for hiding

The `enum` in `DialogItemsDialog` sort of just stores everything needed to pull this off without too much linkage between classes.

Current users upgrading to this should just have everything enabled since I default the value to -1 **but please check this personally if possible**
I am pretty sure this will "just work" as is though. Testing proved it to be true at least.

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Activities/MainActivity.java
+++ b/app/src/main/java/me/ccrama/redditslide/Activities/MainActivity.java
@@ -149,6 +149,7 @@ import me.ccrama.redditslide.Constants;
 import me.ccrama.redditslide.ContentType;
 import me.ccrama.redditslide.FDroid;
 import me.ccrama.redditslide.Fragments.CommentPage;
+import me.ccrama.redditslide.Fragments.DrawerItemsDialog;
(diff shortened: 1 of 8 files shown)


PiwwowPants:
I forgot that my life is always immediately force pushing

PiwwowPants:
The formatter did some wonky stuff I can clean up if you care enough.

PiwwowPants:
I can promise you the only thing I personally did the `MainActivity.java` was add that method and do a call to it. If you want me to remove the formatting for sanity's sake I can. I don't know why this commit of all of them decided to finally respect the code style settings.

ccrama:
Great work, thank you! Now I can finally ignore all my mail

PR #2779, opened 2018-05-19
Allow night mode to follow system in Android M and higher

PiwwowPants:
I made night mode follow the system on Android M and higher.
I additionally hide the options to set your own time if you're on M or higher so the system can just do this for us.

If you'd like the options to retain, I can keep them. But this is without. With, I'd have to add another option to this menu, which is not a big deal but I hate adding options.

Let me know either way and I can make the change if desired.

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/SettingValues.java
+++ b/app/src/main/java/me/ccrama/redditslide/SettingValues.java
@@ -1,6 +1,7 @@
 package me.ccrama.redditslide;
 
 import android.content.SharedPreferences;
+import android.content.res.Configuration;
 
 import net.dean.jraw.models.CommentSort;
 import net.dean.jraw.paginators.Sorting;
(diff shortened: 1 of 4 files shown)


app/src/main/java/me/ccrama/redditslide/SettingValues.java
```diff
@@ -468,8 +470,17 @@ public static TimePeriod getSubmissionTimePeriod(String sub) {
 
 
     public static boolean isNight() {
-        int hour = Calendar.getInstance().get(Calendar.HOUR_OF_DAY);
-        return (hour >= nightStart + 12 || hour < nightEnd) && tabletUI && nightMode;
```
PiwwowPants: I removed the check for `tabletUI`. It felt wrong, but if it is not, let me know and I can put it back.

PiwwowPants:
I think the build erroneously canceled

PiwwowPants:
Tested through an entire day. Works great.

PiwwowPants:
Just pushed a bad commit message description. Same change.

ccrama:
Same as #2781 :D

ccrama:
Beautiful, thank you! Now just waiting on Google to put in a system night mode this will pair nicely

Issue #2775, opened 2018-05-19
Auto-rotation setting ignored while viewing post comments or in subreddit


EthanArmbrust:
Slide version: dbb58dc198ba5cc371a3f7057fbf96fa7d2d9b58
Android version: 8.1.0 and P beta

Happens on commit dbb58dc198ba5cc371a3f7057fbf96fa7d2d9b58 but not previous ones.  
When viewing a post's comments or while browsing a single subreddit, the screen will rotate into landscape view even when auto-rotate is disabled by the OS.

ccrama:
I believe this one is actually a P beta issue, will wait for a future update to see if Google has fixed this one!

EthanArmbrust:
This was also happening in 8.1.0 as well

PiwwowPants:
Thanks for testing on multiple OSs. I had told @ccrama I thought this was an Android P issue when I first saw it so we brushed it off. Should be fixed in my PR.

EthanArmbrust:
Awesome. Thanks!

PR #2780, opened 2018-05-20
Remove screenOrientation setting from CommentsScreen in AndroidManifest

PiwwowPants:
Fixes issue https://github.com/ccrama/Slide/issues/2775

I am absolutely floored no one noticed this. This is actually not an Android P issue, but CommentScreen did not respect orientation lock, only the sensor.

Removing this manifest resolved it on all API levels


ccrama:
Oh wow. Thanks for the fix!

ccrama:
I may or may not have added that to test something

PR #2781, opened 2018-05-20
Rename the SettingValues.tabletUi variable to SettingValues.isPro

PiwwowPants:
I had a lot of confusion about the purpose of this variable until Alex explained it indicates the user has Pro installed.

This variable definitely needs a rename or it will get missed (like I missed it in my Night Mode PR)

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Activities/MainActivity.java
+++ b/app/src/main/java/me/ccrama/redditslide/Activities/MainActivity.java
@@ -511,8 +511,8 @@ public class MainActivity extends BaseActivity
             } else {
                 inflater.inflate(R.menu.menu_subreddit_overview, menu);
             }
-            //Only show the "Share Slide" menu item if the user doesn't have Pro installed
-            if (SettingValues.tabletUI) {
+            // Hide the "Share Slide" menu if the user has Pro installed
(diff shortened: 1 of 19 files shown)


app/src/main/java/me/ccrama/redditslide/Activities/MainActivity.java
```diff
@@ -510,8 +510,8 @@ public boolean onCreateOptionsMenu(Menu menu) {
             } else {
                 inflater.inflate(R.menu.menu_subreddit_overview, menu);
             }
-            //Only show the "Share Slide" menu item if the user doesn't have Pro installed
-            if (SettingValues.tabletUI) {
+            // Hide the "Share Slide" menu if the user has Pro installed
```
PiwwowPants: I also changed this comment to indicate what the next block of code is about to do, not what it will not do if it fails.

PiwwowPants:
Apparently I missed some somehow even though I just refactored with IntelliJ. Maybe I didn't catch the GPlay version.

ccrama:
Seems like the entire Reddit class is a conflict @PiwwowPants 

You broke it!?!?

ccrama:
Last remnant of old naming

Issue #2783, opened 2018-05-21
Content Filters don't catch v.redd.it


ccrama:

PiwwowPants:
It is also not catching youtube links in any form for some reason.
Breakpointing in `PostMatch.doesMatch` is showing youtube get through.

Tensounder54:
This issue is being closed as it is resolved in PR #2784.

MetalNeo:
I am experiencing this issue with v.reddit and youtu.be links not being filtered properly.

Android: 11
Slide: v6.7

PR #2784, opened 2018-05-21
Catch v.redd.it links in content filter

PiwwowPants:
Do not require videoPlugin to catch YouTube links in content filter

Fixes https://github.com/ccrama/Slide/issues/2783

v.redd.it links seem to match as gifs in ContentType. If you'd rather these be videos, we should move in both places. For now, I just included it in the GIF block for PostMatch since it matches ContentType

Additionally it seems the lack of the videoPlugin would cause YouTube links to not get caught in the filter.

Additionally additionally I made it so all Tumblr urls are checked, including *.tumblr.com. Before tumblr.com had to be the start of the URL and would get through the filter

EDIT: Seems you specified these are in fact gifs [on Reddit](https://www.reddit.com/r/slideforreddit/comments/8kv0o5/content_filter/dzavdwv)


PiwwowPants:
Closing since you have this in local changes

PR #2786, opened 2018-05-24
Make contest mode header use correct font color instead of hardcoded white

### LINKED ISSUE #2332: Contest mode header unreadable with light theme
I'll post a screenshot later. Here is an example thread. 

https://www.reddit.com/r/legaladvice/comments/5jq9ee/flmiamiupdatebeing_sued_for_having_topless_photos/

timawesomeness:
Fixes #2332


ccrama:
There used to be a reason for this but I don't remember anymore so thank you for fixing this :)

PR #2787, opened 2018-05-26
Moved Sorting logic out to a Util class

PiwwowPants:
No logic changed. All sorting / search time sorting logic was moved out of the Application class and into a Util class.

Additionally specific context was dropped from the methods of the sorting class in exchanged for the god-context of Application since these instances do not need explicit context of each piece of the application.

It's a LOT but the most I can promise you is nothing changed. Call out any specifics if you're worried about anything. I might have insight into what happened.


PiwwowPants:
Closing in exchange for https://github.com/ccrama/Slide/pull/2790

PR #2790, opened 2018-05-26
Condensed logic in Application class

PiwwowPants:
Moved Sorting logic out to a Util class
Refactored notification channels setup
Removed redundant notification channels used for suppressing audio

This is an appendage of PR https://github.com/ccrama/Slide/pull/2787

It got a bit out of hand but I managed to condense a lot of the logic in the Reddit application class down a lot. There is more work to do but this helps a bit at least.

* Sorting logic is moved to its own Util class. That seems to work and should not hurt
* Using the god context in the `Reddit` application instead of activity-centric context passing to the application level where it's not needed
* Refactored the notification channel setup to just do a loop through all the channels instead of discrete logical blocks
* Removed the two notification channels that specifically handled silent notification
    * As a by product of this, I changed how notifications are built to ensures the `notifsound` setting in Slide is still respected over the settings in Oreo.
* Reduced duplicate code in the `Reddit` application level by passing from overloaded methods to their superiors

There is a lot here. I will be pointing out what is important to analyze for you so your life is not awful.

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Notifications/CheckForMail.java
+++ b/app/src/main/java/me/ccrama/redditslide/Notifications/CheckForMail.java
@@ -14,8 +14,8 @@ import android.content.Intent;
 import android.content.res.Resources;
 import android.net.Uri;
 import android.os.AsyncTask;
-import android.support.v4.app.NotificationManagerCompat;
 import android.support.v4.app.NotificationCompat;
+import android.support.v4.app.NotificationManagerCompat;
 import android.text.Html;
 
 import net.dean.jraw.models.Message;
@@ -78,7 +78,7 @@ public class CheckForMail extends BroadcastReceiver {
             Resources res = c.getResources();
             if (messages != null && !messages.isEmpty()) {
                 Collections.reverse(messages);
-                if (Reddit.isPackageInstalled(c, "com.teslacoilsw.notifier")) {
+                if (Reddit.isPackageInstalled("com.teslacoilsw.notifier")) {
                     try {
 
                         ContentValues cv = new ContentValues();
@@ -129,17 +129,15 @@ public class CheckForMail extends BroadcastReceiver {
                         }
                     }
 
-                    Notification notification =
-                            new NotificationCompat.Builder(c).setContentIntent(intent)
+                    NotificationCompat.Builder builder =
+                            new NotificationCompat.Builder(c, Reddit.CHANNEL_MAIL).setContentIntent(
+                                    intent)
                                     .setSmallIcon(R.drawable.notif)
                                     .setTicker(
                                             res.getQuantityString(R.plurals.mail_notification_title,
                                                     amount, amount))
                                     .setWhen(System.currentTimeMillis())
                                     .setAutoCancel(true)
-                                    .setChannelId(
-                                            SettingValues.notifSound ? Reddit.CHANNEL_MAIL_SOUND
-                                                    : Reddit.CHANNEL_MAIL)
                                     .setContentTitle(
                                             res.getQuantityString(R.plurals.mail_notification_title,
                                                     amount, amount))
@@ -149,8 +147,11 @@ public class CheckForMail extends BroadcastReceiver {
                                     .setGroupAlertBehavior(NotificationCompat.GROUP_ALERT_SUMMARY)
 
                                     .addAction(R.drawable.ic_check_all_black,
-                                            c.getString(R.string.mail_mark_read), readPI)
-                                    .build();
+                                            c.getString(R.string.mail_mark_read), readPI);
+                    if (!SettingValues.notifSound) {
+                        builder.setSound(null);
+                    }
+                    Notification notification = builder.build();
 
                     notificationManager.notify(0, notification);
                 }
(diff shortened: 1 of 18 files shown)


app/src/main/java/me/ccrama/redditslide/Reddit.java
```diff
@@ -761,197 +550,88 @@ public void doMainStuff() {
...
+                        new NotificationChannel(notificationTriple.getLeft(),
+                                notificationTriple.getMiddle(), notificationTriple.getRight());
                 notificationChannel.enableLights(true);
-                notificationChannel.setSound(null, null);
-                notificationChannel.setLightColor(Palette.getColor(""));
-                notificationChannel.setShowBadge(false);
+                notificationChannel.setShowBadge(
+                        notificationTriple.getRight() == NotificationManager.IMPORTANCE_HIGH);
+                notificationChannel.setLightColor(
+                        notificationTriple.getLeft().contains("MODMAIL") ? getResources().getColor(
```
PiwwowPants: I had to string-check the notification channel string since modmail seems to get its own color. This should not be a big deal as long as no one decides to change those strings.

app/src/main/java/me/ccrama/redditslide/Notifications/CheckForMail.java
```diff
@@ -200,26 +201,26 @@ public void onPostExecute(List<Message> messages) {
...
                                                         m.getDataNode().get("body_html").asText())))
                                         .setStyle(notiStyle)
                                         .setGroup("MESSAGES")
                                         .setGroupAlertBehavior(NotificationCompat.GROUP_ALERT_SUMMARY)
                                         .addAction(R.drawable.ic_check_all_black,
-                                                c.getString(R.string.mail_mark_read), readPISingle)
-                                        .build();
+                                                c.getString(R.string.mail_mark_read), readPISingle);
+                        if (!SettingValues.notifSound) {
+                            builder.setSound(null);
```
PiwwowPants: Instead of straight building off the builder to the notification, each notification build is created then I check this setting value to decide if we should suppress audio. This should function without needing to use discrete channels for suppressing audio

app/src/main/java/me/ccrama/redditslide/Notifications/CheckForMail.java
```diff
@@ -200,26 +201,26 @@ public void onPostExecute(List<Message> messages) {
...
-                                new NotificationCompat.Builder(c).setContentIntent(openPi)
+                        NotificationCompat.Builder builder = new NotificationCompat.Builder(c,
+                                Reddit.CHANNEL_MAIL).setContentIntent(openPi)
                                         .setSmallIcon(R.drawable.notif)
                                         .setTicker(res.getQuantityString(
                                                 R.plurals.mail_notification_title, 1, 1))
                                         .setWhen(System.currentTimeMillis())
                                         .setAutoCancel(true)
                                         .setContentTitle(contentTitle)
-                                        .setChannelId(
```
PiwwowPants: I dropped all of these since the builder method takes in the channel you're building on now. No need to chain with this anymore.

app/src/main/java/me/ccrama/redditslide/Reddit.java
```diff
@@ -142,44 +130,16 @@ public static void forceRestart(Context context) {
         ProcessPhoenix.triggerRebirth(context, new Intent(context, MainActivity.class));
     }
 
-    public static void forceRestart(Context c, boolean forceLoadScreen) {
```
PiwwowPants: I dropped this since the boolean was never actually used. If you passed in either true or false, it still did the same thing. I moved the boolean to the superior method that this was overloading and now you just have to pass in either true or false, which is fine.

app/src/main/java/me/ccrama/redditslide/util/SortingUtil.java
```diff
@@ -0,0 +1,212 @@
...
+import java.util.HashMap;
+import java.util.Locale;
+import java.util.Map;
+
+import me.ccrama.redditslide.Activities.Search;
+import me.ccrama.redditslide.ColorPreferences;
+import me.ccrama.redditslide.R;
+import me.ccrama.redditslide.Reddit;
+
+public class SortingUtil {
```
PiwwowPants: I am pretty sure I touched no logic in any of these methods. I just straight copy and pasted out and replaced usages from `Reddit.` to `SortingUtil.`

I did include usages of the god context where the activity context was not necessary.

PiwwowPants:
Notably, even though I removed the channels, they did not remove from my Settings screen in Oreo until I re-installed the app fully.
Not sure why. Maybe Oreo retains deleted channels? Might just be "how it is"?

Either way, the deleted channels are no longer used in any way.

Nxt3:
Dude, you're awesome.

PiwwowPants:
I had to force push again to fix a comment 😢 
I called it "quartet" instead of "triple"

app/src/main/java/me/ccrama/redditslide/util/SortingUtil.java
ccrama: I really like this! Have no idea what possessed me to make this part of Reddit

ccrama:
Looks great, logic is sound just moved around :)

Issue #2789, opened 2018-05-26
When making a new post, text fields are treated as login fields by Oreo autofill


najodleglejszy:
Slide version: 5.8.6
Android version: 8.0

Whenever I create a new post in Slide, all text fields in all submission types (selftext/image/link) are detected by password manager. which shows autofill option on each of them. The only exception is the URL field in Link submission type. Here's a screenshot of Bitwarden offering filling credentials into Title field, but the same thing is happening with Subreddit name and Text. 

![screenshot_20180526-201209](https://user-images.githubusercontent.com/8598559/40579168-40c0e27e-6111-11e8-8fd2-237de5bf06f1.png)

Bitwarden uses Android Oreo's autofill service, so I suspect that other password managers that use it instead of Accessibility service are doing the same thing.
I'm experiencing the issue on Sony Xperia XZ2 Compact, but I've also had it on other phones running Android Oreo.

ccrama:
@PiwwowPants  I believe we can fix it by adding this method to the base activity

`    @Override
    public void onCreate(Bundle savedInstanceState) {
       super.onCreate(savedInstanceState);
       disableAutoFill();
    }

    @TargetApi(Build.VERSION_CODES.O)
    private void disableAutoFill() { 
        getWindow().getDecorView().setImportantForAutofill(View.IMPORTANT_FOR_AUTOFILL_NO_EXCLUDE_DESCENDANTS);
    }
`

PiwwowPants:
@ccrama is there no where in the app where we want to have autofill happening? Would this make autofill in the OAuth view break, assuming it opens internally in the app as a WebView?
Things I could probably just check instead of asking.

ccrama:
We could just override the method in the login subclass, should be the only place we might want it!

Tensounder54:
Issue resolved in PR #2792

PR #2792, opened 2018-05-28
Marked app as unimportant for autofill

PiwwowPants:
Quick and dirty fix for https://github.com/ccrama/Slide/issues/2789

EDIT: I changed how this is implemented so now un-importance is set on the `BaseActivity` and we can override this anywhere up the chain like in `Login` for views that may need to utilize the Autofill API.

Now nothing native should get the Autofill prompt since we're not doing that in the app anywhere natively.

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Activities/BaseActivity.java
+++ b/app/src/main/java/me/ccrama/redditslide/Activities/BaseActivity.java
@@ -1,10 +1,10 @@
 package me.ccrama.redditslide.Activities;
 
+import android.annotation.TargetApi;
 import android.app.ActivityManager;
 import android.content.res.Configuration;
 import android.graphics.Bitmap;
 import android.graphics.BitmapFactory;
-import android.graphics.drawable.BitmapDrawable;
 import android.nfc.NdefMessage;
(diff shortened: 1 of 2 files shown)


app/src/main/res/layout/activity_submit.xml
```diff
@@ -1,10 +1,11 @@
...
-    android:layout_width="match_parent"
-    android:layout_height="match_parent"
-    android:background="?attr/activity_background"
-    android:orientation="vertical">
+                xmlns:app="http://schemas.android.com/apk/res-auto"
+                android:layout_width="match_parent"
+                android:layout_height="match_parent"
+                android:background="?attr/activity_background"
+                android:orientation="vertical"
+                android:importantForAutofill="noExcludeDescendants">
```
PiwwowPants: This will mark the view and all descendants as unimportant for autofill and prevent heuristics from trying to pry out the data for autofill.

PiwwowPants:
"bup bup bup I'm @ccrama, Piwp please do this the right way instead of how you did it bup bup bup"

ccrama:
Love ya @PiwwowPants

PR #2793, opened 2018-05-28
Better support RTL layouts

PiwwowPants:
Better support pre-API 17 layouts
Delete unused layouts

This is big. HUGE.
So giant I would not want to ever be on the receiving end.

I walked the linter to add support for RTL (previously, nothing worked) and pre-API 17 (previously layouts were wonky). This did not resolve ALL issues but RTL looks fully usable now and Ice Cream Sandwich is not a disaster artist.

I am PAINSTAKINGLY looking at all the layouts possible in LTR to make sure I did not break LTR modes but they all look fine still after some programatic fixes.

A small list of what happened here:
* Changed usages of `Gravity.RIGHT` to `Gravity.END` (and LEFT to START if it existed)
* Added some logic in `CreateCardView` to also handle END and START on top of LEFT and RIGHT to fully support both RTL and LTR
* Added usages of START and END in layouts (left and right were not removed)
* Added usages of LEFT and RIGHT in layouts (pre-API 17 hell)
* Changed usages of "right" on gravity in layouts to "end" (and left to start if it existed)

And surprisingly that is it.

I will run with this build for a while to make sure I'm not some pleb that messed up. It's up to @ccrama to decide if this is worth a merge.

Make sure you ignore whitespace changes on the diff to cut it down by around 3000 lines.


app/src/main/res/values/donottranslate.xml
```diff
@@ -1,7 +1,5 @@
 <resources>
 
-    <string name="hello_world">Hello world!</string>
```
PiwwowPants: You never removed this? What a pleb.

app/src/main/res/values/donottranslate.xml
```diff
@@ -16,7 +14,6 @@
     <string name="submission_properties_seperator_comments">"•"</string>
 
     <string name="submission_properties_seperator">"  •  "</string>
-    <string name="general_pro">Slide for Reddit Pro</string>
```
PiwwowPants: The linter said it doesn't care about this and you never used it. It also said "wow way to go never using this @ccrama "

app/src/main/java/me/ccrama/redditslide/Activities/MultiredditOverview.java
```diff
@@ -211,7 +211,7 @@ public void onClick(@NonNull MaterialDialog materialDialog,
                 return true;
 
             case R.id.subs:
-                ((DrawerLayout) findViewById(R.id.drawer_layout)).openDrawer(Gravity.RIGHT);
+                ((DrawerLayout) findViewById(R.id.drawer_layout)).openDrawer(Gravity.END);
```
PiwwowPants: I changed these from RIGHT to END anywhere they existed. Nothing should change for LTR users but RTL users aren't left in the dust anymore

app/src/main/java/me/ccrama/redditslide/Views/CreateCardView.java
```diff
@@ -413,19 +413,27 @@ public boolean onLongClick(View v) {
             RelativeLayout.LayoutParams layoutParams = (RelativeLayout.LayoutParams) v.findViewById(R.id.inside).getLayoutParams();
 
             if (!SettingValues.actionbarVisible && !SettingValues.actionbarTap) {
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN_MR1) {
+                    picParams.addRule(RelativeLayout.START_OF, R.id.secondMenu);
```
PiwwowPants: These bits are the only real logical pieces I added to handle both START/END and LEFT/RIGHT. This was the core bug with RTL/LTR support

app/src/main/res/layout/activity_reader.xml
```diff
@@ -26,7 +26,7 @@
                 android:elevation="8dp"
                 android:minHeight="@dimen/standard_toolbar_height"
                 android:theme="@style/Theme.AppCompat"
-                app:layout_scrollFlags="scroll|enterAlways"></android.support.v7.widget.Toolbar>
```
PiwwowPants: Oh yeah I did this in ONE spot too because I saw the linter complaining with my own eyes. Nothing changed. It just won't be angry you had an empty node.
This happens in other spots too. I just only fixed it here for some reason.

PiwwowPants:
The settings for switching thumbnails is still called "Right-aligned thumbnail" which is confusing and incorrect for RTL. Not a big deal but a thing to potentially change in the future with some more generic wording ("Switch thumbnail position" maybe?).

PiwwowPants:
You wanted it so here it is. I ran a full code cleanup and import optimization against the entire code base. I also resolved low hanging linter issues that should in theory improve performance and fixed a bug. I'll note that bug fix.

app/build.gradle
```diff
@@ -74,7 +74,7 @@ robototextview {
 }
 
 ext {
-    supportLibVersion = '27.1.0'
+    supportLibVersion = '27.1.1'
```
PiwwowPants: The linter was very mad about this because we had conflicting imports from another dependency using 27.1.1 transitively while we were on 27.1.0.

app/src/main/java/me/ccrama/redditslide/Activities/Album.java
```diff
@@ -60,7 +61,7 @@
     private int         adapterPosition;
 
     @Override
-    public void onFolderSelection(FolderChooserDialogCreate dialog, File folder) {
+    public void onFolderSelection(@NonNull FolderChooserDialogCreate dialog, @NonNull File folder) {
```
PiwwowPants: @NonNull and @Nullable annotations all the way down everywhere.

app/src/main/java/me/ccrama/redditslide/Activities/Submit.java
```diff
@@ -585,12 +585,16 @@ public File createFile(Uri uri, @NonNull Context context) {
             String type = resolver.getType(uri);
             String extension;
 
-            if ("image/png".equals(type)) {
-                extension = ".gif";
-            } else if ("image/png".equals(type)) {
-                extension = ".png";
-            } else {
-                extension = ".jpg";
+            switch (type) {
```
PiwwowPants: this switch statement (used in three places) was the bug fix.
`image/png` was being caught twice. The first catch set the extension as a `.gif` erroneously causing `pngs` to save as `gifs` and `gifs` to end up as `jpgs` or just never save. Depending how this was used.

app/src/main/res/values/donottranslate.xml
PiwwowPants: I didn't even press "delete" on any resources. The linter just did it because you didn't use these strings. Up to you if you want them to stick around.

app/src/main/java/me/ccrama/redditslide/Views/ToolbarColorizeHelper.java
```diff
@@ -122,7 +130,8 @@ public void onGlobalLayout() {
         });
     }
 
-    private static void removeOnGlobalLayoutListener(View v, ViewTreeObserver.OnGlobalLayoutListener listener) {
+    private static void removeOnGlobalLayoutListener(View v,
+            ViewTreeObserver.OnGlobalLayoutListener listener) {
         if (Build.VERSION.SDK_INT < Build.VERSION_CODES.JELLY_BEAN) {
```
PiwwowPants: The diff apparently does not care that this is part of the diff, but in these easy cases, I just call the more appropriate method for pre-Jelly-Bean since these were just syntax changes and not actual API changes.

app/src/main/java/me/ccrama/redditslide/Activities/SettingsSynccit.java
```diff
@@ -49,23 +49,28 @@ public void onCreate(Bundle savedInstanceState) {
...
+
+                                            e.putString(SettingValues.SYNCCIT_NAME,
+                                                    SettingValues.synccitName);
+                                            e.putString(SettingValues.SYNCCIT_AUTH,
+                                                    SettingValues.synccitAuth);
+                                            e.apply();
+                                            name.setText(SettingValues.synccitName);
+                                            auth.setText(SettingValues.synccitAuth);
+                                            SynccitRead.visitedIds.removeAll(
+                                                    Collections.singleton("16noez"));
```
PiwwowPants: I feel like I did this and the diff is bad but if I did not, in cases where we only added a single item to a collection, it was translated to a `Collections.singleton()` call which is more performant

PR #2795, opened 2018-05-28
Support fullscreen mode for the internal browser

JaimeLennox:
Extends the WebChromeClient to support custom views to allow fullscreen
options to be used correctly. This is particularly useful for embedded
videos within webpages.


ccrama:
Great work, thank you for the PR!

Issue #2707, opened 2018-04-03
Globe icon in Reader Mode should open the default browser of the device instead of a WebView


armandojrod:
Slide version: 5.8.5

To have the best of both worlds, reader mode for simple articles and a complete browser if you need more.

Alexendoo:
Closed by #2796

PR #2796, opened 2018-05-29
Fixed reader mode and custom tabs not being respected

PiwwowPants:
Converted settings for browser types into an int
Added linkHandlingType enum
Moved reader mode out to a discrete option
Allow reader mode "open external" option to respect selected browser type
Moved intent extras for internal browsers to LinkUtil class

A byproduct of this change adds the feature request https://github.com/ccrama/Slide/issues/2707

I'll do the run down as always:
* The core of this issue was that `ReaderMode` and custom tabs were not respected after my change for link handling. 
    * I blamed the discrete boolean handling so I moved this out to an enum with an int for the `SettingValues`
* After this refactor, it was an opportune time to move `ReaderMode` out of the types of browser handling and into its own discrete option
    * This option appears below the link handling types. When on, all links with open in `ReaderMode`.
    * When the globe is pressed, the user's link handling type is respected
* I moved some constants for intent extras shared between `ReaderMode` and `Website` into `LinkUtil` since that felt more appropriate
* I consolidated the `LinkUtil.openUrl` logic to avoid duplication of code by calling the superior from the overloaded method
* I created a new method for handling theme-ing internal browsers (reader and internal) to avoid mistakes where one is changed and the other is not

As a result, users' defined link handling types will no longer be the same after this change is pushed. The only value that will continue to be respected will be `ReaderMode` since I decided to use the original preference value for this.

Lots of annoying auto-cleanup. But that is pretty much par for the course by now.

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Activities/LiveThread.java
+++ b/app/src/main/java/me/ccrama/redditslide/Activities/LiveThread.java
@@ -22,7 +22,6 @@ import android.webkit.WebChromeClient;
 import android.webkit.WebView;
 import android.widget.HorizontalScrollView;
 import android.widget.ImageView;
-import android.widget.RelativeLayout;
 import android.widget.TextView;
 
 import com.afollestad.materialdialogs.AlertDialogWrapper;
@@ -60,6 +59,7 @@ import me.ccrama.redditslide.Views.CommentOverflow;
 import me.ccrama.redditslide.Views.SidebarLayout;
 import me.ccrama.redditslide.Visuals.Palette;
 import me.ccrama.redditslide.util.HttpUtil;
+import me.ccrama.redditslide.util.LinkUtil;
 import me.ccrama.redditslide.util.LogUtil;
 import me.ccrama.redditslide.util.SubmissionParser;
 import me.ccrama.redditslide.util.TwitterObject;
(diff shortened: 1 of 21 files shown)


app/src/main/java/me/ccrama/redditslide/Activities/LiveThread.java
```diff
@@ -203,62 +203,85 @@ protected Void doInBackground(Void... params) {
                         com.neovisionaries.ws.client.WebSocket ws = new WebSocketFactory().createSocket(thread.getWebsocketUrl());
                         ws.addListener(new WebSocketListener() {
                             @Override
-                            public void onStateChanged(com.neovisionaries.ws.client.WebSocket websocket, WebSocketState newState) throws Exception {
+                            public void onStateChanged(
```
PiwwowPants: Yep. Reformat decided that none of these need to throw an exception... I didn't touch these directly but it IS accurate that they do not need that notation anymore...  *sigh*

app/src/main/java/me/ccrama/redditslide/util/LinkUtil.java
```diff
@@ -107,37 +114,37 @@ public static void openCustomTab(@NonNull String url, int color,
     }
 
     public static void openUrl(@NonNull String url, int color, @NonNull Activity contextActivity,
-            int adapterPosition, Submission submission) {
-        if (!SettingValues.web) {
-            // External browser
+            @Nullable Integer adapterPosition, @Nullable Submission submission) {
+        if (!(contextActivity instanceof ReaderMode) && ((SettingValues.readerMode
```
PiwwowPants: Analyze this entire condition for deciding to open `ReaderMode` and let me know if it's right.

PiwwowPants: "It's not! You'll never enter reader mode if you have night mode on!"
Thanks, other Piwp.

PiwwowPants: This entire condition may be valuable to move somewhere where it's more reliable to call out to since it's a double check.
But now I am just talking to myself so that's up to you @ccrama

PiwwowPants:
Pretty sure I broke the build by force pushing too fast mid-build. Lets see...

ccrama:
Not going to lie, this is a pretty large change that I do not *completely* understand, but since you've been the link handling master I will merge this for an alpha test soon ;)

PR #2806, opened 2018-06-15
Fix MP4 saving not working for cached GIFs

timawesomeness:
As reported [here](https://www.reddit.com/r/slideforreddit/comments/8r7xnm/issue_downloading_gifs/).

Also removed duplicate toast that's already shown by `saveGif`

I think this is the right way to fix this?


ccrama:
Great find, thank you! That fix looks perfect

PR #2809, opened 2018-06-25
Added support for native spoiler markdown

PiwwowPants:
[Request from Reddit](https://www.reddit.com/r/slideforreddit/comments/8tlyb9/could_slide_please_get_support_for_this_new/)

I added support for marking the new "native" spoiler markdown as spoilers in the app. The regex checks out.
Some changes did occur with the logic for marking as spoilers but it should not break classic markdown. **If someone could hook me up with a link to text the non-native spoiler code or test it on your own that'd help.**

Also a clean up because everything is awful.

Please verify the regex is not wrong and should work / test this as well on top of my testing to ensure I did not mess it up. It'd be a bit bad to flub this one.

Additionally I added some very minimal tests for the patterns. They should at least fail on changes but they aren't exhaustive by any means and just do a very general check

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/SpoilerRobotoTextView.java
+++ b/app/src/main/java/me/ccrama/redditslide/SpoilerRobotoTextView.java
@@ -1,38 +1,17 @@
 package me.ccrama.redditslide;
 
 import android.app.Activity;
-import android.content.ClipData;
+import android.content.*;
 import android.content.ClipboardManager;
-import android.content.Context;
-import android.content.ContextWrapper;
-import android.content.DialogInterface;
-import android.content.Intent;
 import android.content.res.TypedArray;
-import android.graphics.Bitmap;
-import android.graphics.BitmapFactory;
-import android.graphics.Color;
-import android.graphics.PorterDuff;
-import android.graphics.Typeface;
+import android.graphics.*;
 import android.graphics.drawable.Drawable;
 import android.net.Uri;
 import android.os.AsyncTask;
 import android.os.Environment;
 import android.support.v4.content.ContextCompat;
-import android.text.Html;
-import android.text.Spannable;
-import android.text.SpannableStringBuilder;
-import android.text.Spanned;
-import android.text.TextPaint;
-import android.text.style.BackgroundColorSpan;
-import android.text.style.CharacterStyle;
-import android.text.style.ForegroundColorSpan;
-import android.text.style.ImageSpan;
-import android.text.style.QuoteSpan;
-import android.text.style.RelativeSizeSpan;
-import android.text.style.StrikethroughSpan;
-import android.text.style.StyleSpan;
-import android.text.style.TypefaceSpan;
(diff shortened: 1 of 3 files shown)


app/src/main/java/me/ccrama/redditslide/SpoilerRobotoTextView.java
```diff
@@ -83,7 +59,9 @@
     private              List<Integer>        storedSpoilerStarts = new ArrayList<>();
     private              List<Integer>        storedSpoilerEnds   = new ArrayList<>();
     private static final Pattern              htmlSpoilerPattern  =
-            Pattern.compile("<a href=\"([#/](?:spoiler|sp|s))\">([^<]*)</a>");
+            Pattern.compile("<a.*href=\"[#/](?:spoiler|sp|s)\".*?>(.*?)</a>");
```
Alexendoo: It's important that the pattern does not track past the current tag, the `.*` after a will match greedily, for example the following will be caught in a single match: `<a href="#s">one</a> <a href="#s">two</a>`

A negative character class `[^>]*` would be the appropriate replacement since just making it lazy wouldn't handle `<a href="https://not-a-spoiler.com">one<a/> <a href="#s">two</a>` correctly

app/src/main/java/me/ccrama/redditslide/SpoilerRobotoTextView.java
```diff
@@ -83,7 +59,9 @@
...
+    private static final Pattern nativeSpoilerPattern =
+            Pattern.compile("<span.*class=.*?md-spoiler-text.*?>(.*?)</span>");
```
Alexendoo: Same problem with the `.*` in this pattern, the `.*?` between `class` and `md-spoiler-text` may also over-track if reddit has or introduces (I don't think it does yet?) any other formatting using spans

ccrama: @Alexendoo I think you are correct about spans, but they could definitely add something without notice and cause problems

app/src/main/java/me/ccrama/redditslide/SpoilerRobotoTextView.java
```diff
@@ -212,14 +190,18 @@ private void replaceQuoteSpans(Spannable spannable) {
...
-        while (htmlSpoilerMatcher.find()) {
-            String newPiece = htmlSpoilerMatcher.group();
+        html = formatSpoilers(html, htmlSpoilerPattern.matcher(html));
+        html = formatSpoilers(html, nativeSpoilerPattern.matcher(html));
+        return html;
+    }
+
+    private String formatSpoilers(String html, Matcher matcher) {
+        while (matcher.find()) {
+            String text = matcher.group();
```
Alexendoo: Should this be `.group(1)`?

ccrama:
Awesome, thank you very much @PiwwowPants @Alexendoo !

PR #2815, opened 2018-07-08
Display Wikis as HTML in a WebView

PiwwowPants:
- I removed any native implementation for the Wiki rendering. **If there is something still hanging out let me know so I can remove it**
- I replaced everything in the layout for the WikiPage with a single webView that just gets the HTML data from the API loaded into it
    - I added a cover over the WebView that hides when the page finishes loading. This is to avoid a white flash before the styles are loaded in
- The customCss is set on the Wiki activity statically then accessed from the WikiPage. This is the prevent every fragment from recreating it since it only needs created once per subreddit as it is now
    - This CSS only modifies the background color, font color, and idle link text to match the default themes
    - There is a commented out method that makes tables horizontally scrollable in the view but that conflicted with the ViewPager so I commented it out and added a TODO
- I redid the AsyncTask on the WikiPage because it was in the way too often. It's a static inner class now that does the same thing but "safer"
- I overrode loading Urls on the WikiPage WebView
    - If the link out of the WikiPage is to itself, it calls a listener up to the Wiki activity to select the correct fragment
    - Otherwise is just tries to open it however it needs to via the app or externally based on the user's browser preferences
    - If it tries to open an embedded WikiPage that does not exist, I show a dialog matching Reddit's error for that situation

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Fragments/WikiPage.java
+++ b/app/src/main/java/me/ccrama/redditslide/Fragments/WikiPage.java
@@ -2,34 +2,57 @@ package me.ccrama.redditslide.Fragments;
 
 import android.os.AsyncTask;
 import android.os.Bundle;
+import android.support.annotation.NonNull;
 import android.support.v4.app.Fragment;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
-
-import java.util.List;
-
+import android.webkit.WebView;
+import android.webkit.WebViewClient;
 import me.ccrama.redditslide.Activities.Wiki;
 import me.ccrama.redditslide.Constants;
+import me.ccrama.redditslide.OpenRedditLink;
 import me.ccrama.redditslide.R;
-import me.ccrama.redditslide.SpoilerRobotoTextView;
-import me.ccrama.redditslide.Views.CommentOverflow;
 import me.ccrama.redditslide.Views.GeneralSwipeRefreshLayout;
 import me.ccrama.redditslide.Visuals.Palette;
-import me.ccrama.redditslide.util.SubmissionParser;
+import net.dean.jraw.managers.WikiManager;
+import org.apache.commons.text.StringEscapeUtils;
+
+import java.lang.ref.WeakReference;
+
 
 public class WikiPage extends Fragment {
     private String title;
     private String subreddit;
+    private String wikiUrl;
+
+    private WikiPageListener listener;
+
+    private WebView webView;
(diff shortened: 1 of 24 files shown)


ccrama:
This is very awesome! Would it be possible to add a fade in so the white isn't blinding on the initial load?

Thanks again :)

app/src/main/java/me/ccrama/redditslide/Fragments/WikiPage.java
```diff
@@ -38,66 +61,48 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle sa
...
+                    String pagePiece = url.toLowerCase().replace(wikiUrl.toLowerCase(), "")
+                            .split("/")[0]
+                            .split("#")[0];
+                    listener.embeddedWikiLinkClicked(pagePiece);
+                } else {
+                    OpenRedditLink.openUrl(getContext(), url, true);
                 }
-                return null;
+                return true;
             }
```
Alexendoo: Think I found a way to get rid of the white flashes, adding:

```java
            @Override
            public void onPageStarted(WebView view, String url, Bitmap favicon) {
                webView.setBackgroundColor(Color.TRANSPARENT);
                super.onPageStarted(view, url, favicon);
            }
```

took care of it for me. Might be more performant to set the correct background colour rather than transparent but I couldn't figure out which class to use for that. Seems to be some duplicated things between Palette.java and ColorPreferences.java

app/src/main/java/me/ccrama/redditslide/Fragments/WikiPage.java
```diff
@@ -38,66 +61,48 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle sa
...
-            } else {
-                commentOverflow.setViews(blocks.subList(startIndex, blocks.size()), subredditName);
-            }
-        } else {
-            commentOverflow.removeAllViews();
-        }
+    private void onDomRetrieved(String dom) {
+        webView.loadDataWithBaseURL(
+                wikiUrl,
+                "<head>".concat(Wiki.getGlobalCustomCss()).concat("</head>").concat(StringEscapeUtils.unescapeHtml4(dom)),
```
Alexendoo: For the super huge wikis, the majority of the execution time is spent in unescapeHtml4, moving the call into the WikiAsyncTask doInBackground seems to completely alleviate the main thread blocking

ccrama:
Looks great, amazing work! We're lucky to have you helping out with Slide :)

Issue #2816, opened 2018-07-14
Embedded links that reference sub-page wikis do not work


PiwwowPants:
With the new WebView-based Wiki implementation, if a link on the Will references a sub-page (e.g. "Recommended/Books") the new page will not be loaded. 

This is most likely due to how the WikiPage title is pulled from the URL, as I believe it doesn't care about anything after a forward slash. 

Can be tested with the /r/lifeisstrange Wiki and selecting "Books" from the recommended page. This action should open the "Recommended/Books" page.

Tensounder54:
This issue is resolved in a later version of Slide and needs to be closed.

PR #2819, opened 2018-07-21
Improved WikiPage implementation

PiwwowPants:
Corrected issue where Wiki sub-pages would not open from links
Allow table and code elements to be horizontally scrollable in WebView

Fixes issue https://github.com/ccrama/Slide/issues/2816

This change makes it so subpages (e.g. "recommended/books") in Wikis are opened correctly when clicking a link within the HTML

Additionally this change reformats `table` and `code` elements in the HTML to allow them to overflow on the x-axis be scrollable. This is done by injecting some JS that determines if the user touched a `table` or `code` dependent element and disables the ViewPager paging in the native code until their pointers are released from the screen.

Pros: You can scroll `table` and `code` blocks appropriately. This means the entire WebView does not scroll off the x-axis requiring panning before paging. Additionally we do not have to reformat these elements which would lead to them being rendered in an unexpected or unreadable way.
Additionally the way I implemented this allows it to extend to any element easily (written for `table` and `code` was just dropped in last minute and works fine)

Cons: You absolutely cannot page the pager on these elements regardless if they are actually overflowing the view. That logic in the JS would be pretty annoying to write and maintain and I feel it is an edge case of a problem. In cases the entire WikiPage is a `table` or `code` block this would just entirely kill paging, but again I believe that to be an edge case. The biggest problem is if a user is unaware why it is happening.

Let me know how you feel. and if you have any better ideas in your head for this.

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Activities/Wiki.java
+++ b/app/src/main/java/me/ccrama/redditslide/Activities/Wiki.java
@@ -10,12 +10,12 @@ import android.support.design.widget.TabLayout;
 import android.support.v4.app.Fragment;
 import android.support.v4.app.FragmentManager;
 import android.support.v4.app.FragmentStatePagerAdapter;
-import android.support.v4.view.ViewPager;
 import com.afollestad.materialdialogs.AlertDialogWrapper;
 import me.ccrama.redditslide.Authentication;
 import me.ccrama.redditslide.ColorPreferences;
 import me.ccrama.redditslide.Fragments.WikiPage;
 import me.ccrama.redditslide.R;
+import me.ccrama.redditslide.Views.ToggleSwipeViewPager;
 import me.ccrama.redditslide.Visuals.Palette;
 import me.ccrama.redditslide.util.LogUtil;
 import net.dean.jraw.managers.WikiManager;
@@ -32,12 +32,13 @@ public class Wiki extends BaseActivityAnim implements WikiPage.WikiPageListener 
     public static final String EXTRA_PAGE = "page";
 
     private TabLayout tabs;
-    private ViewPager pager;
+    private ToggleSwipeViewPager pager;
     private String subreddit;
     private Wiki.OverviewPagerAdapter adapter;
     private List<String> pages;
     private String page;
     private static String globalCustomCss;
+    private static String globalCustomJavaScript;
 
     @Override
     public void onCreate(Bundle savedInstance) {
@@ -51,6 +52,7 @@ public class Wiki extends BaseActivityAnim implements WikiPage.WikiPageListener 
 
         applyColorTheme(subreddit);
         createCustomCss();
+        createCustomJavaScript();
         setContentView(R.layout.activity_slidetabs);
         setupSubredditAppBar(R.id.toolbar, "/r/" + subreddit + " wiki", true, subreddit);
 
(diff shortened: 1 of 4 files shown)


app/src/main/java/me/ccrama/redditslide/Fragments/WikiPage.java
```diff
@@ -70,12 +73,15 @@ public void run() {
...
+        webView.getSettings().setJavaScriptEnabled(true);
+        webView.addJavascriptInterface(new WikiPageJavaScriptInterface(), "Slide");
+
         webView.setWebViewClient(new WebViewClient() {
             @Override
             public boolean shouldOverrideUrlLoading(WebView view, String url) {
                 if (url.toLowerCase().startsWith(wikiUrl.toLowerCase()) && listener != null) {
                     String pagePiece = url.toLowerCase().replace(wikiUrl.toLowerCase(), "")
-                            .split("/")[0]
+                            .split("\\?")[0]
```
PiwwowPants: I will mention that this is now presumptive. I just assume everything after the base Reddit Wiki URL is the page identifier up until any parameters. I don't know if I can do anything better than that.

app/src/main/java/me/ccrama/redditslide/Activities/Wiki.java
```diff
@@ -83,12 +85,26 @@ private void createCustomCss() {
...
         customCssBuilder.append("</style>");
         globalCustomCss = customCssBuilder.toString();
     }
 
+    private void createCustomJavaScript() {
+        globalCustomJavaScript = "<script type=\"text/javascript\">" +
+                "window.addEventListener('touchstart', function onSlideUserTouch(e) {" +
+                "var element = e.target;" +
+                "while(element) {" +
+                "if(element.tagName && (element.tagName.toLowerCase() === 'table' || element.tagName.toLowerCase() === 'code')) {" +
```
PiwwowPants: It was pointed out that I call `*.toLowerCase()` twice, extraneously. I should probably move the `tagName` to a variable to avoid that and make the operation marginally faster.

app/src/main/java/me/ccrama/redditslide/Activities/Wiki.java
```diff
@@ -83,12 +85,26 @@ private void createCustomCss() {
                 .append("color: ".concat(getHexFromColorInt(ta.getColor(2, Color.BLUE))).concat(";"))
                 .append("; }");
         ta.recycle();
-        // TODO: Implement horizontal scrolling to Wiki table views
-        //customCssBuilder.append("table { display: block; overflow-x: auto; white-space: nowrap; }");
+        customCssBuilder.append("table, code { display: block; overflow-x: scroll; }");
+        customCssBuilder.append("table { white-space: nowrap; }");
```
PiwwowPants: I believe this to have been a poor choice since it causes table rows to extend on single lines infinitely. This should be removed or a max width should be added to the going style

ccrama:
> Cons: You absolutely cannot page the pager on these elements regardless if they are actually overflowing the view. That logic in the JS would be pretty annoying to write and maintain and I feel it is an edge case of a problem. In cases the entire WikiPage is a table or code block this would just entirely kill paging, but again I believe that to be an edge case. The biggest problem is if a user is unaware why it is happening.

I think that is totally understandable from a use perspective, that is what code blocks/tables do in Slide comments/submissions anyway

ccrama:
Amazing work as always, thank you!

PR #2831, opened 2018-08-05
Allow night mode to be set manually when devices support automatic night mode

PiwwowPants:
This change reverses my removal of the manual night mode options when a device support automatic night mode.

It replaces the boolean for night mode with an int that is based on the ordinals of values in a new NightModeState enum that tell us if the user has night mode disabled, set manually, or being followed automatically. I included a forward migration strategy inline and a TODO to remove that later (which I hope everyone reads.)

Not fully tested yet but seems to work. I recommend just testing this yourself manually to see if everything is triggering correctly.

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/SettingValues.java
+++ b/app/src/main/java/me/ccrama/redditslide/SettingValues.java
@@ -2,18 +2,16 @@ package me.ccrama.redditslide;
 
 import android.content.SharedPreferences;
 import android.content.res.Configuration;
-
+import me.ccrama.redditslide.Fragments.SettingsHandlingFragment;
+import me.ccrama.redditslide.Views.CreateCardView;
+import me.ccrama.redditslide.Visuals.Palette;
+import me.ccrama.redditslide.util.SortingUtil;
 import net.dean.jraw.models.CommentSort;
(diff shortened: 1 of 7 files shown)


app/src/main/java/me/ccrama/redditslide/SettingValues.java
```diff
@@ -513,4 +514,12 @@ public static boolean hasSort(String subreddit) {
     public enum ColorMatchingMode {
         ALWAYS_MATCH, MATCH_EXTERNALLY
     }
+
+    public enum NightModeState {
+        DISABLED, MANUAL, AUTOMATIC;
```
PiwwowPants: If these get moved around or get out of sync with the resource array, everything breaks. Notably.

app/src/main/res/layout/nightmode.xml
```diff
@@ -23,25 +23,23 @@
 
                 <TextView
                         android:id="@+id/title"
-                        android:layout_width="wrap_content"
+                        android:layout_width="0dp"
```
PiwwowPants: The spinner was not respecting its gravity well so I just made the title consume the rest of the white space that the spinner was not. I don't just suck.

ccrama:
I like it :D

PR #2837, opened 2018-08-13
Set FLAG_ACTIVITY_NEW_TASK on external link intents

PiwwowPants:
Call LinkUtil.openExternally from Website context external button

This is a hopeful fix for external intents crashing the app. I added the NEW_TASK flag to the intents since we're launching externally off of the application context, so the new context can't attach back to us.

Does this even make sense? I'm going to try to vet it out to at least one other person before deciding it works.

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Activities/Website.java
+++ b/app/src/main/java/me/ccrama/redditslide/Activities/Website.java
@@ -4,42 +4,22 @@ import android.content.Intent;
 import android.net.Uri;
 import android.os.Bundle;
 import android.support.v7.widget.Toolbar;
-import android.view.Menu;
-import android.view.MenuInflater;
-import android.view.MenuItem;
-import android.view.View;
-import android.view.ViewGroup;
-import android.view.WindowManager;
-import android.webkit.CookieManager;
-import android.webkit.CookieSyncManager;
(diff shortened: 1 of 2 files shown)


ccrama:
We used to do that, but removed it because it caused a lot of issues 😅

ccrama:
How would that tag be causing the crashing issues?

PiwwowPants:
My hunch is that because I made a change to call the intent off of the application context, some devices/launchers are crashing when the intent tries to attach itself to the previous context given. If I use baseContext, that may be resolved since it's an activity context, but I'm not completely sure yet.

Flagging it for a new task should let the called intent open very discretely in the external app we call to open.

I'm only 20% sure about this until I get something back from the tester.

PiwwowPants:
Both of the people I sent an APK off to seem to claim that this change fixed their external link handling issues @ccrama. 

The other factor here is I'm sending off noGPlayDebug builds instead of the production APK. I have no reason to believe that that would affect this change. 

I think based on that I think it'd be reasonable to say this resolved it. Can you recall the issues this caused in the past?

ccrama:
That is a very interesting idea, I think you are right! There should be no reason a debug build should affect it. 

I will also push a new tag with this commit so F-Droid gets a new build ASAP!

PR #2864, opened 2018-10-05
Replace IRC link in README

Alexendoo:
I missed this one when changing the badge over


ccrama:
Thanks for the fix @Alexendoo !

PR #2866, opened 2018-10-07
Add support for /best sorting endpoint

christarazi:
Fixes #2803 

Depends on this PR first:  ccrama/JRAW#12

PR #2867, opened 2018-10-07
Do not show "Best" option in default settings

christarazi:
I realized after pushing #2866 that the default settings under general incorrectly display "Best" as an option (since it is only valid for the frontpage). This fixes that.

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Activities/MultiredditOverview.java
+++ b/app/src/main/java/me/ccrama/redditslide/Activities/MultiredditOverview.java
@@ -469,6 +469,10 @@ public class MultiredditOverview extends BaseActivityAnim {
                         .toLowerCase(Locale.ENGLISH);
         final Spannable[] base = SortingUtil.getSortingSpannables("multi" + id);
         for (Spannable s : base) {
+            // Do not add option for "Best" in any subreddit except for the frontpage.
+            if (s.toString().equals(getString(R.string.sorting_best))) {
+                continue;
+            }
             MenuItem m = popup.getMenu().add(s);
         }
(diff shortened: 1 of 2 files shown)


christarazi:
@ccrama Have you had a chance to take a look at these changes?

timawesomeness:
It also shouldn't be shown when a subreddit is opened as an overlay, and on profiles

Alexendoo:
It seems like removing it from where it _shouldn't_ be seems a little error prone&mdash;is there a way instead to only explicitly add it to frontpage?

Alexendoo:
That sounds like a good plan. I think that would allow fixing the other thing I noticed too, that best was getting added to the bottom of the list whereas on reddit desktop it's the top entry

ccrama:
I might be missing something but best sorting works on all subreddits for me? Works on desktop as well

timawesomeness:
It's only a post sort on the home page. Subs just have hot. It's different from the comment sort.

ccrama:
Hmm I don’t think that’s true. On the reddit redesign you can set sorting to hot on any subreddit. Is there any official word on how that works?

ccrama:
* best. Sorry, replying from email

timawesomeness:
You're not the first person I've seen say that, but it's not that way for me, I only see hot in subreddits.

Regardless i'm pretty sure it's lying and still sorting by hot because best doesn't make any sense as anything other than a home page sort, because it's built around personalizing the home page by prioritizing posts from subs you interact.

ccrama:
@christarazi It appears that an old commit removed best from sorting in the MainActivity, and I have just removed it from the SingleSubreddit view as well. Merging now, thank you for the PR!

PR #2874, opened 2018-10-17
Add support for Reddit Silver and Platinum awards

### LINKED ISSUE #2876: Display the Silver/Platinum gildings alongside Gold
https://www.reddit.com/coins

Quick example of what it looks like on desktop:
image

Currently, only gold is shown:
image

Possible addition to this: an option to have gilding icon/text scaled down to the same size as the flair text, since with up to three different icons and counters it can get big quickly.

Slide version: 5.9 a2
Android version: 8.1.0

timawesomeness:
Depends on ccrama/JRAW#13.

Idk about the two new icons, I based them on the current gold icon and the colors from reddit's new official icons. They look like [this](https://i.imgur.com/9I97gjo.png) ([on a light background](https://i.imgur.com/5AI9VBF.png), [on a gray background](https://i.imgur.com/s1gOiyd.png)). If you want them modified in any way before merging I can do that.

Also adds award icons for modqueue comments, whereas previously only a confusing unlabeled number was shown.

Also fixes an old obscure bug that I encountered preventing comment navigation by gilded comments from working if not logged in.

closes #2876


ccrama:
Glad Reddit is complicating it even more :D

Great job with the icons, looks great :)

PR #2875, opened 2018-10-18
Fix several "peek content" crashes

timawesomeness:
Fixes:
* Crash on peeking a post that is inaccessible
* Crash on peeking a comment that is inaccessible
* Crash on peeking a shadowbanned or deleted user account
* Crash on peeking a permanently suspended user account


ccrama:
Great work!

PR #2879, opened 2018-10-25
Fix build failure

Alexendoo:
Updates JRAW to the latest as required by #2866

Fixes a weird bug with the respository ordering, by putting google() at the top it stops the `Could not find lint-gradle-api.jar` error from appearing


ccrama:
Glad that fixes it! I keep getting spammed by Jenkins :P

PR #2880, opened 2018-10-29
Don't show sticky action on non-self comments

timawesomeness:
Reddit doesn't let you sticky comments you didn't make. Prevents a crash reported [here](https://www.reddit.com/r/slideforreddit/comments/9shntl/when_you_try_to_sticky_a_comment_that_isnt_yours/).


ccrama:
Oh interesting, I have tried doing this in the past because I thought it was possible. Thanks for the fix!

Issue #2798, opened 2018-05-30
Auto correct doesn't work in "title" field when creating a new post


incIined:
Slide 5.8.6
S9+ TouchWiz Oreo, 2/1/18 patch level

![](https://i.imgur.com/gPU7Tt9.jpg) 

Basically, it gives autocorrect suggestions but instead of replacing the words, just leaves it as is.  The bold on the word you see on my suggestion bar never appears when typing in the title field.

Tensounder54:
This issue is resolved in a later version of Slide and needs to be closed.

ryder55:
hi, can you please provide me link to an issue where this is fixed? when i tried it two weeks ago it wasnt working.

Tensounder54:
The issue is resolved in Slide version 5.9-alpha2. Although it appears that your PR was merged so...

Tensounder54:
Issue resolved in PR #2884

PR #2884, opened 2018-11-01
Add "textAutoCorrect" value to "title" field on Submit post screen

ryder55:
Hi, I've pushed changes regarding issue #2798


ccrama:
While I was holding off on doing that, a lot of users have requested it and I think it's a worthy addition now. Thanks for the PR!

PR #2890, opened 2018-11-10
Don't show suspended dialog for logged-in user

### LINKED ISSUE #2711: Minor annoyance: it's impossible to view your own account if you're temporarily suspended
It pops up with the "This account is suspended" dialog, and when you press OK it leaves your profile. It'd be nice if that didn't dismiss your profile when trying to view your own temporarily suspended account.

timawesomeness:
Logged in users can see their profile when they're suspended, so there's no reason to show the dialog and prevent profile access.

Closes #2711


ccrama:
Good find, thank you!

PR #2891, opened 2018-11-10
Show others' multis more than once

### LINKED ISSUE #2888: Other users' multireddits are only shown the first time you try to view them
Resulting in a nice empty page with lots of crashy buttons the second and subsequent times.

timawesomeness:
This looks like something that was forgotten when this code was refactored at some point.

Fixes #2888


ccrama:
Definitely what happened there, thanks for finding that!

PR #2893, opened 2018-11-11
Fix crash when going to the profile backing subreddit of a suspended/deleted account

timawesomeness:
When an account with the new profile is suspended or deleted, the subreddit backing that profile (/r/u_username) is put in a weird state where it still exists, but has no content and minimal info. Slide expects certain properties, like `over18`, to not be null, so when they are null it crashes. This makes Slide consider those subreddits to be inaccessible so it doesn't crash.


ccrama:
Interesting, thank you for the fix!

PR #2894, opened 2018-11-11
Fix post notifications

### LINKED ISSUE #2841: Notifications for posts/Subreddits not coming through.
Report from [this Reddit post](https://www.reddit.com/r/slideforreddit/comments/92qfkx/subscribed_post_notifications_dont_notify_anymore/)

I've noticed this issue as well. When subscribing to subreddit updates, I don't get any notifications. I do seem to get notifications about messages coming in to me though.

I can't verify the report from Reddit easily, but it may be beneficial for others to verify that these notifications are no longer being pushed through.

timawesomeness:
Post notifications were broken because reddit removed timestamp search back in April. This reverts the change that switched Slide to use timestamp search. Fixes #2841.

Also adds some spacing around separators in post notifications so they look better.

Also makes non-subscribed subs removable from post notifications.

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Fragments/SettingsGeneralFragment.java
+++ b/app/src/main/java/me/ccrama/redditslide/Fragments/SettingsGeneralFragment.java
@@ -26,6 +26,7 @@ import com.afollestad.materialdialogs.DialogAction;
 import com.afollestad.materialdialogs.MaterialDialog;
 import com.rey.material.widget.Slider;
 
+import me.ccrama.redditslide.*;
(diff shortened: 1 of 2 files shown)


ccrama:
Honestly it's probably a good idea to not use Reddit search anyway, it typically sucks. But we should also make sure the user notif interval is less than an hour, otherwise they won't see anything new (if it's set to 5 hours, for example)

timawesomeness:
Apparently that setTimePeriod(hour) didn't actually do anything, so I removed it. It should work with any notification interval.

ccrama:
Great. Thank you!

PR #2896, opened 2018-11-12
Fix the FAB reply dialog

### LINKED ISSUE #2873: [FAB Reply Module] Content Area does not have a vertical scroll overflow
https://i.imgur.com/7IfY8jI.png

It looks like the module that pops up when using the reply FAB does not have a vertical scroll overflow (which the edit module does).  This causes long comments (like when making a table) to hide the buttons to submit the reply.

Even longer bodies of text also hides the formatting toolbar.

Slide version: 5.9alpha2
Android version: 9.0

timawesomeness:
Fixes #2873, fixes [this](https://www.reddit.com/r/slideforreddit/comments/9mztd6/is_anyone_else_having_this_problem/)

Switches back to the edit_comment layout and adds the account switcher to that.


ccrama:
One of the most reported bugs, thank you for the fix!

PR #2898, opened 2018-11-13
Support displaying OC tags, let users mark own posts as spoilers

timawesomeness:
* Add support for displaying reddit's newish OC tags
* Let users mark their own posts as spoilers


ccrama:
Did not know that tag existed, need to do this on iOS as well! Thanks again!

PR #2899, opened 2018-11-13
Add search as a FAB option

timawesomeness:
As requested [here](https://www.reddit.com/r/slideforreddit/comments/982f6z/search_as_fab/)


ccrama:
This is something I was going to bring over from iOS as well :) Thank you!

Issue #2915, opened 2018-11-28
gfycat gifs don't open


bradii2:
Slide version: 5.9-alpha2
Android version: 9

the-eater:
Just about to open this :')

Example link: https://www.reddit.com/comments/a0y20n

Screenshot: 
![screenshot_20181128-085257_slide](https://user-images.githubusercontent.com/1922630/49136813-13e7e100-f2eb-11e8-89bd-3082c7e6ae44.png)

bradii2:
Yeah, I think maybe gyfcat updated their website in a way that broke slide just earlier today. It's the only thing I can think of

the-eater:
This happend before: #2600 may be fixed in a moment, may be the end of it working

PR #2917, opened 2018-11-28
Update gfycat url

JaimeLennox:
I need GIFs in my life! :scream: 

The old endpoint 404s; I've tested this endpoint and it appears to work fine.

Fixes: https://github.com/ccrama/Slide/issues/2915


ccrama:
Thank you for the fix! Gfycat is putting the old api back up for a bit of time so we can get an update out

Nxt3:
The old endpoint still isn't being used and Slide 404's.

ccrama:
Gfycat is the one that 404’s, Slide is using the best thing it can as of the last commit

Nxt3:
Oops, yeah I meant Gfycat.

My point is that we need this update :P

ccrama:
Gfycat is going to proxy the old api to the new one for me for the time being, today sometime! But this does push the update to be more important

PR #2918, opened 2018-11-28
Fix post sticky crash

timawesomeness:
Apparently you can't sticky a removed post (even though you can remove a stickied post).

As reported [here](https://www.reddit.com/r/slideforreddit/comments/a14f3t/mods_should_not_be_able_to_pin_a_removed_post/).


ccrama:
Strange issue, thanks for the fix!

PR #2924, opened 2018-12-01
Revert temporary fix and properly fix sidebar/sub search/app bar/subredditview crash

timawesomeness:


ccrama:
Thank you very much for looking into this! You saved the day

PR #2959, opened 2019-01-19
Text editor improvements

### LINKED ISSUE #2712: Change author icon in editor items to use an image instead of a TextView
https://github.com/ccrama/Slide/blob/master/app/src/main/res/layout/editor_items.xml

Use: https://materialdesignicons.com/icon/at

### LINKED ISSUE #2948: Insert image into selected text
I've noticed when I select text and click insert url, it correctly works, but not when I click insert image.

timawesomeness:
* Better handling of newlines when selecting and formatting text
* Correct handling of newlines when quoting text
* Properly unescape quoted text
* Hide soft keyboard when opening quote dialog
* Image link insertion with selected text like the link button does (closes #2948)
* Make image insertion dialog not dismissable on tap outside to prevent accidental image link loss
* Add spoiler button
* Switch to an image for mention button so it tints properly (closes #2712)
* Make mention button insert at the cursor instead of the end


ccrama:
Great changes as always, thank you!

Issue #2791, opened 2018-05-27
Bad performance on really long threads


alexsmithfanning:
Slide 5.8.6

LineageOS 14.1

Moto X Pure

This might be due to my aging phone, but really long threads cause lag when performing any action but scrolling. Scrolling is still smooth, but when you go to upvote or downvote the animation chops and takes a couple seconds before the tap registers.

This only happens when you're really far down in the thread.

This has been happening for as long as I can remember, even back to the earlier days. It might not even be fixable.

Nxt3:
Yeah this has been a thing since forever. Even on my Pixel 2XL

alexsmithfanning:
@ccrama I sent you a report via the "send feedback" button because the lag got so bad that it eventually froze.

alexsmithfanning:
@Nxt3 good to know it's not just that my phone is getting old.

ccrama:
@Nxt3 I've never experienced this on many devices, is it possibly a setting you and @alexsmithfanning have enabled in Settings > Comments?

alexsmithfanning:
@ccrama try [this thread](https://www.reddit.com/r/AskReddit/comments/8lj4fn/what_small_thing_should_be_illegal_because_it/) and keep loading more commends. It should start to get laggy and eventually cripple itself.

ccrama:
@alexsmithfanning I loaded 10k more and have no lag opening menus or voting, even on the bottom comments. Mind posting your comment settings screen?

Nxt3:
Here are my comment settings

![](https://i.imgur.com/dVoFVL8.png)

![](https://i.imgur.com/BkrBHSe.png)

ccrama:
Ah, the issue is fully collapse comments, that allows hundreds more views to stay on the screen at a time, pushing the limits of the RecyclerView and how we do comments. Unsure if there is a way to remedy that

PiwwowPants:
Even with child comments expanded by default I made it 10k down plus the children and everything was fine on my end. Assuming this is a RecyclerView that makes sense and shouldn't hit this issue naturally. I'll try to copy settings exactly and see.

alexsmithfanning:
Hmm. I'm not noticing it after Slide crashed and I had to restart it. I'm thinking it has to be open for a while, and multiple long threads have to be view and exited before the issue appears. I'll keep experimenting.

PiwwowPants:
[...]

PR #2960, opened 2019-01-20
Performance improvements

timawesomeness:
This is nowhere near all the performance improvements Slide needs, but these didn't involve rewriting too much and provide a noticeable improvement

* Improved listing loading time
  * optimized sqlite queries (~50% total query time reduction, -2 sec loading time on my device)
  * optimized filtering and content typing (~500% filtering time improvement, -130ms per doesMatch call on my device)
* Improved profile listing loading time 
  * optimized sqlite query (same optimization as above) (~20% total query time reduction)
  * moved sqlite queries and filtering to background to prevent UI from freezing for several seconds
* Improved performance while scrolling with collapsed comments (helps with #2791 but doesn't totally fix it - other part of that issue is with setTextHtml/getBlocks in setViews in onBindViewHolder in CommentAdapter)
  * switched hidden from ArrayList to HashSet (up to ~350% faster getRealPosition calls)


Tensounder54:
Woop, woop!

Issue #2963, opened 2019-01-24
Show a proper error when trying to add more than 100 subs to a multireddit


Soberat:
Slide version:6.0.1-3
Android version: 8.1
Right now when you try to add a subreddit to an already full multireddit (100 subs) you get a generic error, and I think it would be better if the app checked the size of the multi first before even sending an API request

RubenNL:
Shouldn't this issue be closed?

PR #2971, opened 2019-02-07
Show a proper error when trying to add more than 100 subs to a multireddit

MrLhama:
Fix for issue #2963
Changed the way the error message is built and added the error text in the strings.xml file


ccrama:
I was not aware of that limitation, good find!

PR #2996, opened 2019-03-23
Use jsoup to infer Reader mode encodings

### LINKED ISSUE #2995: Reader mode should fall back onto UTF-8 rather than ISO-8859-1
Slide version: Slide v6.1 (F-Droid)
Android version: 9 (Lineage OS 16.0-20190317-NIGHTLY-oneplus2)
Issue Point in code: https://github.com/ccrama/Slide/blob/master/app/src/main/java/me/ccrama/redditslide/Activities/ReaderMode.java#L112

When opening links that dont declare their charset in the http headers the reader decides to decode them as ISO-8859-1. This is mainly a issue as [since HTML 4.0](https://www.w3.org/TR/html4/intro/intro.html#h-2.3.1) UTF-8 has been the defult charset of HTML document. also not all sites define their charset in the headers, the site that trigged this issue has it defined with this `<?xml version="1.0" encoding="UTF-8"?>` and this `<meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />`

Screenshot of issue: Screenshot of issue

Reddit Content with link that triggers issue: https://reddit.com/r/engrish/comments/b4emtx/dunno_if_this_counts_but_its_still_weird/ej6do2n?context=3

Site that triggers issue : https://artpark.or.jp/sansaku/artist/hongousin/no24.html

Site Headers: Site Headers

Alexendoo:
We already have it as a dependency for readability itself so there's no extra cost here. It infers based on the content type header or the http-equiv within the HTML itself, falling back to UTF-8 if none of those things exist.

Fixes #2995

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Activities/ReaderMode.java
+++ b/app/src/main/java/me/ccrama/redditslide/Activities/ReaderMode.java
@@ -14,17 +14,9 @@ import com.afollestad.materialdialogs.AlertDialogWrapper;
 import com.wuman.jreadability.Readability;
 
 import org.apache.commons.text.StringEscapeUtils;
+import org.jsoup.Connection;
+import org.jsoup.Jsoup;
 import org.jsoup.nodes.Document;
-import org.jsoup.nodes.Element;
-
-import java.io.File;
-import java.io.IOException;
-import java.io.InputStreamReader;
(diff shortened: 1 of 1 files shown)


ccrama:
Thank you for the PR! Should fix some of those weird page renderings

PR #2997, opened 2019-03-24
Decelerate submissions animations

Scrumplex:
This PR adds the LinearOutSlowInInterpolator to a few animations in the app, that did not have any interpolator. This makes the animations feel smoother than before.

Added dependencies:
- `com.android.support:interpolator`


ccrama:
Small change but makes a big difference, thank you!

PR #3024, opened 2019-06-05
ahkuma/minor_fixes [DON'T MERGE][WIP]

ahkuma:
Minor fixes for the app and improvements. Work in progress.

**Improvements (Subjective)**
- Use square edge corner instead of rounded corner for search settings. I think it looks better.

**In Progress**
- Automatically hide navigation and status bar when phone is tilted while video is playing.

**Future Improvements**
- Add a button to allow for full screen horizontal without enabling it in system settings.
- Fix issue where the sunglasses icon shows after rotate, which only disappears after you tap on it again.
- Fix issue where if "I am 18 and want to see NSFW content" is enabled, no post in an NSFW sub is loaded.
- Automatically set "Hide all NSFW photos" and "Hide NSFW previews in collections" to false when "I am 18 .... " is set to true and vice versa.

Anyone who reads this, please tell me what you think. Also @ccrama


Tensounder54:
Looks good. There are a few relatively easy High Priority issues in the issue tracker what you court take a look at if you find the time.

ahkuma:
@Tensounder54 

I will definitely look at those, another thing is I feel like the app needs an UI overhaul and a better looking default color scheme. A theme list and a custom option would be nice in my opinion, but I think for now bug fixes and maintenance stuff (library updates etc) should be prioritized.

Issue #2943, opened 2018-12-27
Going into a post and opening the keyboard, then swiping back to the main screen won't close the keyboard automatically. 


incIined:
Slide v6.0.1-3
Android 9.0 (Dec security patch)

Tensounder54:
Thanks for letting us know about this. We will look into it when we get a chance. At the moment it's pretty low priority thought.

> ↳T*54*

Issue #2941, opened 2018-12-27
Replying to a comment and then scrolling down causes the keyboard to stay up, but will not add the words typed while scrolled down


incIined:
Slide v6.0.1-3
Pixel 3 (Android 9.0, Dec security patch)

This is a bit confusing so screen recording below:
https://streamable.com/bmt0p

This also applies for replies from inbox and normal comments - scroll down far enough and the keyboard will stay up autocorrect will work, but nothing will appear in the comment box. (Also in the case of inbox to comments, clicking reply won't shift the screen down to have the comment box in the viewable area of the screen.)

eaglz05:
I have this problem too.

anonymous-user:
I can reproduce this too. It seems like the comment box loses input focus when the user scrolls or the text box is not in view. 

I have a Pixel 2 (Android 9.0, December patch) and Slide 6.0.1-3.

https://photos.app.goo.gl/kAWdzQQNi1ZCAahu9

CreamyCookie:
Sadly the issue still exists in 6.2.1

eaglz05:
@Tensounder54 I can still reproduce this.

timawesomeness:
@CreamyCookie @eaglz05 you're correct that it's not fixed, I'll look into it further

Issue #2931, opened 2018-12-07
When replying to comments, the textbox doesn't automatically lift up over the keyboard


SinTan1729:
Slide version: 6.0.1-3
Android version: 9.0

Tensounder54:
What make and model of phone are you running Slide on?
Are you signed in when this issue occurs? 
Are you using the version from FDroid or the Google Play Store?

SinTan1729:
@Tensounder54 I'm using a Mi A1, although on a custom ROM (Nitrogen OS).
Yes, I'm signed in.
I'm using the FDroid version.

Tensounder54:
OK cool. I'm going to mark this as a bug and we will take a look at this ASAP.

PR #3029, opened 2019-06-29
Partial fix for keyboard not focusing on Pie

timawesomeness:
Pie does something fucky with focus and the keyboard - as best I can tell, the user taps the reply button, we focus the textbox, then the keyboard steals focus before anything can happen so the textbox doesn't move up or gain real focus. The keyboard *does* give focus back to the app once it's popped up, but Pie isn't handling that focus correctly, it's more like fake focus.

This is a hacky, gross, really absolutely horrible workaround, and doesn't always work, but it fixes the most annoying instances of the issue.

May or may not fix the following: #2931, #2941, #2943


CreamyCookie:
Sadly, it doesn't fix the issue for me. Thanks though!

PR #3048, opened 2019-08-07
Use Android Uri's

### LINKED ISSUE #3047: Opening subreddit link with query string and no trailing slash fails
Slide version: 6.2
Android version: 9

**BUG**

When opening a subreddit URL, if the link contains any query string e.g. https://www.reddit.com/r/gifs?fakeQuery then the subreddit will load, but the app shows a failure message and does not allow you to view the subreddit. Suspiciously, this does _not_ happen when linking directly to a post, and appears to occur only when the URL does not have a final ```/```, e.g.:

https://reddit.com/r/gifs?queryString - FAIL
https://reddit.com/r/gifs/?queryString - PASS

**REPRODUCTION**

Allow Slide to open reddit URLs by default. Click a link to a subreddit (e.g. https://www.reddit.com/r/programming?utm_medium=android_app ). The subreddit will load but an error will show.

 **SCREENSHOT**

Screenshot_20190807-130109_Slide

Alexendoo:
This replaces some of the String handling of URLs with `Android.net.Uri`s. Though certainly not all of it, just anything that directly used `formatRedditUrl`

Fixes #3047

I tested a fair amount of URLs, might have missed something though


Alexendoo:
Oops broke the tests, I'll fix that tomorrow

Tensounder54:
@ccrama I don't see why we can't merge this.

Issue #3051, opened 2019-08-13
Misuse of AsyncTask


pangeneral:
According to our research, there are misuses about the following classes:
1. `me.ccrama.redditslide.Vote`
2. `me.ccrama.redditslide.Activities.CommentsScreenSingle.AsyncGetSubredditName` 
3. `me.ccrama.redditslide.Activities.Inbox`
4. `me.ccrama.redditslide.Activities.LiveThread` 

The problems are:
1. There are many annoymous inner AsyncTask classes in `LiveThread` and `Inbox`. They hold strong reference to the GUI element of Activity, which can lead to memory leak when the Activity is destroyed and the AsyncTask did not finish. Similar problem also exists in `Vote` and `AsyncGetSubredditName`.
2. The instances of these AsyncTask classes are not cancelled before the Activity is destroyed, which can lead to the wrong invocation of onPostExecute

I think we can make following changes to fix the misuse problems:
1. I think the GUI-related fields should be wrapped into WeakReference. Take `private View v` in `Vote` as example, it can be changed tp `private WeakReference<view> v`. Besides, use non-anonymous inner static classes to replace these annoymous classes.
2. Add a AsyncTask field in the corresponding Activities which use AsyncTask. These field refer to the annoymous AsyncTask object. Then invoke `cancel()` in the onDestroy() method of Activities.

These are my suggestions above, thank you.

hassan963:
Following changes may need to solve this issue, following @pangeneral advices

1. For anonymous inner AsyncTask classes in `LiveThread` and `Inbox` , make separate AsyncTask classes for them and call them from `LiveThread `and `Inbox`. By this we can make them static class.
2. Call cancel() in onPause() method in Activity.

I am working on this issue, please let me know if this is the correct way. Thank you.

pangeneral:
> Following changes may need to solve this issue, following @pangeneral advices
> 
> 1. For anonymous inner AsyncTask classes in `LiveThread` and `Inbox` , make separate AsyncTask classes for them and call them from `LiveThread `and `Inbox`. By this we can make them static class.
> 2. Call cancel() in onPause() method in Activity.
> 
> I am working on this issue, please let me know if this is the correct way. Thank you.

I think that's correct. Happy to help you :)

PR #3058, opened 2019-08-27
Asynctask issue #3051 solved

hassan963:
Below classes changed and inner `AsyncTask `classes are changed to separate static classes. Also added `WeakReference<>` for context and view. Added `cancel()` method for `AsyncTask` in `onPause()` method in `Activity`.
```
1.     me.ccrama.redditslide.Vote
2.     me.ccrama.redditslide.Activities.CommentsScreenSingle
3.     me.ccrama.redditslide.Activities.Inbox
4.     me.ccrama.redditslide.Activities.LiveThread

```
Separate `AsyncTask `classes are in `Services `package
```
1. me.ccrama.redditslide.Services.CommentScreenTask.java
2. me.ccrama.redditslide.Services.InboxTask.java
3. me.ccrama.redditslide.Services.LiveThreadTask.java

```

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Activities/CommentsScreenSingle.java
+++ b/app/src/main/java/me/ccrama/redditslide/Activities/CommentsScreenSingle.java
@@ -10,6 +10,7 @@ import android.support.v4.app.Fragment;
 import android.support.v4.app.FragmentManager;
 import android.support.v4.app.FragmentStatePagerAdapter;
 import android.support.v4.view.ViewPager;
+import android.util.Log;
 import android.util.TypedValue;
 import android.view.KeyEvent;
 import android.view.ViewGroup;
@@ -31,6 +32,7 @@ import me.ccrama.redditslide.LastComments;
 import me.ccrama.redditslide.Notifications.NotificationJobScheduler;
 import me.ccrama.redditslide.R;
 import me.ccrama.redditslide.Reddit;
+import me.ccrama.redditslide.Services.CommentScreenTask;
 import me.ccrama.redditslide.SettingValues;
 import me.ccrama.redditslide.UserSubscriptions;
 import me.ccrama.redditslide.util.LogUtil;
@@ -49,6 +51,7 @@ public class CommentsScreenSingle extends BaseActivityAnim {
     private String    name;
     private String    context;
     private int       contextNumber;
+    private CommentScreenTask.AsyncGetSubredditName asyncGetSubredditName;
 
     @Override
     protected void onActivityResult(int requestCode, int resultCode, Intent data) {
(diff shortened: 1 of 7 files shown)


app/src/main/java/me/ccrama/redditslide/Vote.java
```diff
@@ -13,19 +13,21 @@
...
 /**
  * Created by ccrama on 9/19/2015.
  */
 public class Vote extends AsyncTask<PublicContribution, Void, Void> {
 
     private final VoteDirection direction;
-    private View v;
-    private Context c;
+    private WeakReference<View> v;
+    private WeakReference<Context> c;
```
Alexendoo: Thanks for the PR @hassan963 

With the adoption of weak refs in this sort of case, what happens to the NPEs if .get() returns null, does it cause a crash?

hassan963: Yes, it will crash. I have checked if the ref is null or not in Vote.java class, so it wont cause any NPEs. But by mistake I did not check on other classes. Below classes
```suggestion
1. me.ccrama.redditslide.Services.CommentScreenTask.java
2. me.ccrama.redditslide.Services.InboxTask.java
3. me.ccrama.redditslide.Services.LiveThreadTask.java
```

app/src/main/java/me/ccrama/redditslide/Activities/CommentsScreenSingle.java
```diff
@@ -209,70 +211,26 @@ public void onPageScrollStateChanged(int state) {
...
 
-        @Override
-        protected void onPostExecute(String s) {
-            subreddit = s;
-            setupAdapter();
+    @Override
+    public void onPause() {
+        super.onPause();
+        if (asyncGetSubredditName != null) {
+            asyncGetSubredditName.cancel(true);
```
Alexendoo: onPause doesn't seem like the right place for this - it would cancel if a user switches apps temporarily, I imagine it should be in onDestroy

hassan963: If we cancel it `onDestroy()` then there is no surety to invoke the method.  What about doing it in `onStop()` method? If you are sure about this then I can change it to `onDestroy()` .

Alexendoo: If the app is killed before onDestroy happens (e.g. After a pause) the whole process goes, which includes any asynctasks. So there's no worry of leaking that way

hassan963: Okay, I get it now. I am going to make those changes for all the activities and then commit those changes.

PR #3060, opened 2019-08-30
Handle CSS and invalid colours in toolbox usersnotes

Alexendoo:
Toolbox comes from the web, so users may use colours that are valid in CSS but not in Android's `Color.parseColor` (for example `orange` is a valid name in CSS but not one `parseColor` accepts).

As it happens exoplayer has a method that can parse most CSS colours, with a larger colour map and support for some other formats. It doesn't look like it has full support of all valid CSS colours, but it's an improvement that doesn't require adding a new dependency.

If the colour fails to parse, it now falls back to grey to prevent crashing

cc @timawesomeness


ccrama:
Interesting, thank you for the improvement!

PR #3066, opened 2019-09-09
Multireddit dropdown list in drawer

philippTheCat:


app/src/main/res/layout/drawer_loggedin.xml
```diff
@@ -407,32 +407,65 @@
         </LinearLayout>
     </RelativeLayout>
 
-    <LinearLayout
+    <RelativeLayout
+        android:id="@+id/multireddits"
```
Alexendoo: Take the `@+id/multi` from the LinearLayout here, you can delete the LinearLayout's id since it doesn't need to be referred to

This is important since you can disable individual drawer items from showing up in the settings. It does this by hiding the `multi` id in this case. Here's what it looks like with multireddits disabled without that

image

You'll have to change `LinearLayout multiReddits = header.findViewById(R.id.multi);` to be a RelativeLayout in MainActivity also

app/src/main/java/me/ccrama/redditslide/Activities/MainActivity.java
```diff
@@ -1529,6 +1523,24 @@ public void onClick(@NonNull MaterialDialog dialog, @NonNull DialogAction which)
                     return true;
                 }
             });
+            LinearLayout multireddits_area = header.findViewById(R.id.multireddits_area);
+
+            multireddits_area.setVisibility(View.GONE);
```
Alexendoo: You can set the view to be initially gone in the XML, slightly preferable to doing it in code. `android:visibility="gone"`

app/src/main/res/layout/multireddit_textview_white.xml
```diff
@@ -0,0 +1,22 @@
...
+        android:textColor="#fff"
+        android:gravity="center_vertical"
+        android:layout_alignParentBottom="true"
+        android:text="multireddit"
+        android:textSize="14sp"
+        />
+
+
+
+</RelativeLayout>
```
Alexendoo: It'd be worth running the android studio formatter on this and `drawer_loggedin` to get the 8 space property indents

app/src/main/java/me/ccrama/redditslide/Activities/MainActivity.java
```diff
@@ -1482,21 +1483,14 @@ public void doDrawer() {
         if (Authentication.isLoggedIn && Authentication.didOnline) {
 
             header = inflater.inflate(R.layout.drawer_loggedin, drawerSubList, false);
+            RelativeLayout multiReddits = header.findViewById(R.id.multi);
             headerMain = header;
             hea = header.findViewById(R.id.back);
 
             drawerSubList.addHeaderView(header, null, false);
             ((TextView) header.findViewById(R.id.name)).setText(Authentication.name);
-            header.findViewById(R.id.multi).setOnClickListener(new OnSingleClickListener() {
```
Alexendoo: With this removed it's not possible to open the regular MultiredditOverview

PR #3103, opened 2019-11-26
Add option to disable image download button

### LINKED ISSUE #3031: Make the Download button in image/video previews optional
Slide version: v6.2
Android version: 9.0

I constantly accidentally touch the download button in image previews and I think it would be a good idea to have this button be optional through a switch in the settings.

michaelPotter:
Fixes #3031


ccrama:
Thank you for the PR, I'll include this in the next update!

Issue #3098, opened 2019-11-20
Doesn't respect metered WiFis


101lols:
The app treats metered WiFis like any other WiFi connection instead of treating them as mobile data.

https://developer.android.com/reference/android/support/v4/net/ConnectivityManagerCompat#isactivenetworkmetered

Tensounder54:
This issue has been resolved and needs to be closed.

PR #3112, opened 2019-12-30
Treat metered WiFi connections as mobile to help with data usage

YusefOuda:
In response to issue #3098


ccrama:
Good change, didn't know it was possible to check for that! Thanks again.

Issue #3078, opened 2019-10-10
Back Button goes to Front Page


sasschary:
When you press the back button while looking at a subreddit you are subscribed to, an exit dialog shows, or the app immediately exits, or the sidebar opens, depending on how you have everything configured. I propose having the back button go back to the front page before performing the current back button behavior. I often find myself trying to use that to go back to the front page, only to be greeted by an exit dialog. This would prevent that from happening, and provide an easier way to return to the frontpage.

sonya323:
Can I work on this issue?

Tensounder54:
@sasschary Thismis the current behavior of the bark button within the app so I'm not sure what you are proposing.

sasschary:
That doesn't seem to be how it works on either of my devices... If I am subscribed to, say, r/AskReddit, and I visit that subreddit (such that it is the one selected on the top bar, not so that it opens in a new window), then press the back button, it asks me if I want to exit Slide. If I go to a subreddit that I am not subscribed to, then it will open like its own, new window on top of everything else, and then, it will go back to the front page when you press the back button. But, that is not how it works for subreddits which you are subscribed to.

PR #3113, opened 2019-12-30
Adds new behavior option for back button to go to first subreddit in …

YusefOuda:
Adds new behavior option for back button to go to first subreddit in the list.

Re: #3078


ccrama:
Thank you for the PR, I'll include this in the next update!

PR #3121, opened 2020-02-12
Synccit supports https

RubenNL:
I have not been able to test this, because i have no experience with developing android apps, but it should work.


Alexendoo:
Seems like it does indeed. Thanks for the PR

PR #3123, opened 2020-02-14
Show subreddit banner on the top of the sidebar (if exists)

YusefOuda:
Example screenshot included to show what it will look like when a banner img exists.

Screenshot_1581647568


ccrama:
Thank you for the PR! This will be included in the next update

Issue #3134, opened 2020-03-07
Save individual files to custom folders


omgwtfbbqpls:
Please, allow us the option to choose where we download/save each individual image. On desktop I have designated folders for different categories (humor, art, info, etc.). On mobile there is no such option, so in the end I'm left with hundreds of images I have to sort manually into my desktop folders (which is really, really annoying).

omgwtfbbqpls:
@Archeidos good shit my man

PR #3144, opened 2020-03-22
Save individual files to custom folders

Archeidos:
Added the another "Save image/MP4 to" item to MediaView that will allow users to manually select their save location for each image they download. Users can still save to their default location using the regular "Save image/MP4" button.

This feature was requested in https://github.com/ccrama/Slide/issues/3134

This is my first open source contribution to a project so please let me know if there is anything that could be optimized or cleaned up. Thanks!


ccrama:
Great work @Archeidos, thank you for the PR! I will be including this in the next Slide update :)

Issue #3119, opened 2020-01-15
FR: Automatically hide read posts server-side


bobobo1618:
I tend to hop between iOS (where I use Apollo) and Android a lot. One feature that Apollo has that I'm yet to see elsewhere is auto-hiding of read posts, so that you never see the same post twice.

I messed around with Apollo to see how it accomplishes this and it appears that it does it by hiding the posts on Reddit, that is, as posts are read in the app, I see them added to https://www.reddit.com/user/<me>/hidden/.

My proposal for how this should work is:

- Keep track of which posts have been read client-side, paired with a boolean of whether they've been hidden server-side.
- Before every fetch of new posts (be it refresh, change subreddit or loading another page), make a batch request to `/api/hide` with the ID of all the read but not hidden posts.
- Fetch the content

I see that this was discussed in https://github.com/ccrama/Slide/issues/2638 but I think the concern about blank pages is resolved, as the filtering is done server-side.

If you can give me some code pointers I'm happy to try implementing this myself.

PR #3146, opened 2020-03-23
Automatically hide read posts on Reddit

bobobo1618:
This addresses #3119.

When a user scrolls past a post, it is added to an in-memory queue of posts waiting to be hidden. Whenever the user performs an action that would load more posts (like switching subreddits or scrolling some more), all of the posts in the queue are hidden server-side and the queue is empty.

This ensures that a user will only ever see the same post once.

I also tried to make the code I was touching a little bit cleaner.

I'm not very experienced with Java so let me know if my code is terrible.

This is intended to be a starting point, I don't expect anything to be merged right away. There isn't even an option to toggle this behaviour yet.


Tensounder54:
So the first thing we'd need to look at befor considering a merge like this is, as you said a toggle. This tech might also be able to be used with Synccit to ensure that seen posts are hidden. It's a good idea but realistically, we'd need to see a toggle before we could consider the PR.

bobobo1618:
(Finally) added a toggle.

PR #3156, opened 2020-04-25
Disambiguate Turkish translations for short durations

hayalci:
's' was used for both seconds and hours, which is catastrophic. Instead, add another letter from "saniye" and "saat" to disambiguate.
'd' for 'minute' is problematic because it can be confused with an untranslated 'day'. Making it use another letter from "dakika" makes it 100% non-ambiguous.

### DIFF
--- a/app/src/main/res/values-tr-rTR/strings.xml
+++ b/app/src/main/res/values-tr-rTR/strings.xml
@@ -791,9 +791,9 @@
     <!-- Time: shorthand durations for example 1 hour → 1h -->
     <string name="time_years_short">%dy</string>
     <string name="time_days_short">%dg</string>
-    <string name="time_hours_short">%ds</string>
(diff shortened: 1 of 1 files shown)


Alexendoo:
Thanks @hayalci, the translations files are generated from the crowdin data (every so often), so you'd have to submit the changes there. Otherwise they get overridden by the update script

I did spot you've made some of the changes on crowdin though, thanks!

Currently it's showing `%dsn` for seconds and `%dst` for hours in crowdin, opposite to how it is in this PR, if that's an accident you should be able to add a new translation and delete the old ones, if not let me know

https://crowdin.com/translate/slide-for-reddit/2/en-tr#q=time+short

hayalci:
Huh, I made a mistake here actually. Benefits of crowdin I guess, showing other translations led me to use correct short form there.

Yeah, it makes sense to drop this, then.

PR #3162, opened 2020-05-04
Fixes 3026: auto-copy URL for old devices

### LINKED ISSUE #3026: Add option to copy url of post or comment to clipboard
Non stock ROMS do not have the copy to clipboard function in the share menu. In any case, it would be nice to have the option to copy the url of a post or comment to clipboard directly in Slide.

goforbg:
Fixes #3026 : Copying URL support for old devices when showing share Intent

Didn't inflate a new item in the Bottom sheet, instead invoked auto-copy URL for older devices support.

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Reddit.java
+++ b/app/src/main/java/me/ccrama/redditslide/Reddit.java
@@ -2,6 +2,8 @@ package me.ccrama.redditslide;
 
 import android.annotation.TargetApi;
 import android.app.*;
+import android.content.ClipData;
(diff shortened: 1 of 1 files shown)


Alexendoo:
I'm not sure it's a good idea to overwrite the users clipboard without a clear indication that's what they expect to happen

goforbg:
Fair enough, looking forward to implement any suggestions for this fix

tobiscode:
Is this related to https://github.com/Haptic-Apps/Slide/issues/3430?

PR #3183, opened 2020-06-10
Handle Subreddit loading error gracefully

ctbur:
When a loading error happens, the `SubmissionsView` would be swapped for an empty view. This PR changes the behavior to show a retry button as shown in the screenshot:

screenshot of the change

I would have preferred it if the button would say: "Retry", instead of: "Would you like to try again?", but this text was already available as a resource with translations.


Tensounder54:
Yeah, I actually really like this. I don't see why this can't be merged but I don't want to do it without @ccrama's permission.

ccrama:
Looks great. Thank you!

Issue #3191, opened 2020-06-22
Option to always show floating action button when scrolling


darshan099:
Slide version: 6.4
Android version: 9

An option to toggle between hiding floating action button and always showing floating action button in the settings.
Since the floating action button is visible only when its scrolled upwards a little, it becomes tedious to perform scrolling upwards continuously to hide posts, especially when we have scrolled past a lot of posts ( with hide post on scroll enabled ).

It will be easier to hide posts without the need to scroll upwards to make floating action button visible.

PR #3192, opened 2020-06-22
Added always show FAB in Settings-General

darshan099:
Fixes issue #3191 : Added a toggle to always show or hide floating action button in Settings->General


ccrama:
Good stuff, thank you for the PR! I might be changing the behavior to ignore the case where the setting is enabled, as it would be calling .show() on a view that is already shown

darshan099:
Yeah changing the behaviour to ignore that case seems correct. Thank you for merging my PR ( it was also my first PR! ).

ccrama:
Thank you for the PR! The update with your new option is live now :)

darshan099:
Thank you so much!

Issue #3194, opened 2020-06-30
Subreddit name when making a new post is too short


LostGhost1:
Slide version: 6.4.324
Android version: 9

My account name is 20 characters long, and Slide won't let me post on my account due to length limit when entering subreddit name. So, I cannot post to u_StalinTheMemeLordBlg
![Screenshot_20200630-194625_Slide](https://user-images.githubusercontent.com/46618828/86153462-65310580-bb0a-11ea-843f-9722690056ed.png)

PR #3195, opened 2020-06-30
Fix issue with posting to profile with 20 character long names

LostGhost1:
Fixes this issue https://github.com/ccrama/Slide/issues/3194


ccrama:
Good find! Thank you for the PR

PR #3197, opened 2020-07-01
Fix crash with %uXXXX encoded chars in toolbox removal reason text

timawesomeness:
https://www.reddit.com/r/slideforreddit/comments/gi8auc/my_app_crashes_when_i_choose_more_than_one/


ccrama:
Thank you for the PR!

PR #3203, opened 2020-07-05
Fix crash when entering offline mode with no cache

### LINKED ISSUE #3199: Offline mode of death
Slide version: 6.4
Android version: 9

0. Make a Slide backup, if your Slide data is important to you.
1. Delete all the offline-cached data (via the sidebar > manage offline content > click all the little X's next to sub names).
2. Go into offline mode.

Result: Slide will consistently crash when starting. I have had two delete all my Slide user data twice because of this bug now, so I'd say this is moderately annoying.

timawesomeness:
Slide crashes when "search in the app bar" is enabled and there's no cached offline content and you enter offline mode.

Fixes #3199


Tensounder54:
OOC, is this to resolve an existing issue listed here on the GitHub? Just wondering, because if it is then I'll mark the issue with the `Will Close On Merge` tag.

ccrama:
Interesting bug, looks like that fixes it! Thank you! Should close #3199

PR #3207, opened 2020-07-11
Support mod locking/unlocking individual comments

timawesomeness:


Tensounder54:
Oh wait, we didn't already support this? Interesting.

timawesomeness:
We already support locking/unlocking whole posts, just not the newer locking individual comments

ccrama:
Just a follow up @timawesomeness did we figure out why JRAW CI was failing for this commit?

timawesomeness:
The JRAW repository's travis config requests jdk7 which AFAIK is unavailable on travis

PR #3221, opened 2020-07-15
Add gradle wrapper validation action

Alexendoo:
Details here: https://github.com/gradle/wrapper-validation-action

It's a solution from Gradle to ensure the jar is always a correct one from them


TacoTheDank:
This is a good idea

PR #3224, opened 2020-07-15
Fixes #3079 where archive notification covers part of the post title

### LINKED ISSUE #3079: "archived post" note is to big, reaches over the post
Slide version 6.2.2
Android version: 9

Title.

I cannot scroll further up. The post title is unreadable.

Screenshot_20191013-131351_Slide

Ullas-Aithal:
This issue was happening because getMeasuredHeight() for the text view doesn't add extra px if the text is multi-line. Followed [this thread](https://stackoverflow.com/questions/41779934/how-is-staticlayout-used-in-android/41779935#41779935) to paint the string and then calculate the actual height of the text in the textview. I've tested this on my phone OnePlus 7 Pro running android 10 for archived post, commented post and combination of these two posts. 
Methods getLineSpacingMultiplier() and getLineSpacingExtra() are available after API 16. So for lower devices, I have set default values.


TacoTheDank:
I happened to be working on AndroidX while you made this.

Sorry about that. I have fixed the conflict.

Issue #2951, opened 2019-01-09
[Feature Request] Provide a setting that will hide or fully collapse all removed/deleted comments with no visible children


abqnm:
Since deleted comments with no children are automatically hidden, I'm just wondering if it would be possible to create a setting option that allows the user to enable automatic hiding (or fully collapsing) of all removed/deleted comments that have child comments that are also removed or deleted. 

This would be especially useful for subs like /r/Science that remove large amounts of comments for being off topic, and can sometimes make it hard to find the comments that aren't removed in the sea of removed comments. So if removed comments that have all the children also removed/deleted could be hidden or fully collapsed automatically, it would make navigating these threads significantly easier.

Thanks in advance!

PR #3246, opened 2020-08-07
Collapse deleted and removed comments by default #2951

Barrett97:
Adds #2951


TacoTheDank:
@ccrama This one should be squashed.

PR #3274, opened 2020-10-01
The lint rolling never stops

TacoTheDank:
Wow, this is fun.

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Fragments/WikiPage.java
+++ b/app/src/main/java/me/ccrama/redditslide/Fragments/WikiPage.java
@@ -113,10 +113,11 @@ public class WikiPage extends Fragment {
     private void onDomRetrieved(String dom) {
         webView.loadDataWithBaseURL(
                 wikiUrl,
-                "<head>".concat(Wiki.getGlobalCustomCss())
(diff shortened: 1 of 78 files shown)


TacoTheDank:
Fixed the end-of-line bullshit that was marking some files as entirely changed.

app/src/main/java/me/ccrama/redditslide/Fragments/WikiPage.java
```diff
@@ -113,10 +113,11 @@ public void onPageFinished(WebView webView, String url) {
     private void onDomRetrieved(String dom) {
         webView.loadDataWithBaseURL(
                 wikiUrl,
-                "<head>".concat(Wiki.getGlobalCustomCss())
-                        .concat(Wiki.getGlobalCustomJavaScript())
-                        .concat("</head>")
-                        .concat(dom),
+                ((("<head>"
```
Alexendoo, line 116: Are the triple parens required? Seems a bit odd

TacoTheDank: Oops, nope they aren't. That was done automatically. There's other ones like this throughout the project, so I think it's fine.

PR #3285, opened 2020-10-27
Add settings to hide awards

Admicos:
This PR adds two settings to hide the display of awards (such as gold) on posts and comments

- Appearance > Post layout > Hide awards
- Appearance > Comments > Hide awards

I don't see any breaking issues with the code, so it should be ready to review. However, I'm worried that the comment setting might effect unwanted areas as I found 5 classes that had similar logic, compared to the single class for posts.


ccrama:
Looks good to me, thank you for the PR! Will include this in the next update

PR #3290, opened 2020-11-14
Fixes notification getting stuck on 'downloading' after a sucessful download

### LINKED ISSUE #2957: Notification still shows downloading image after download is complete
Slide version: 6.0.1-3
Android version: 9.0
Device: Galaxy Note 9

screenshot

budde25:
fixes #2957 

Just a quick fix for the the downloading notification bug I have been experience recently, it doesn't always happen but I noticed that it only happened if it was my first time downloading that image.
I'm not sure if it was a race condition but I was unable to reproduce it after adding that line.

Thanks!

### DIFF
--- a/app/src/main/java/me/ccrama/redditslide/Notifications/ImageDownloadNotificationService.java
+++ b/app/src/main/java/me/ccrama/redditslide/Notifications/ImageDownloadNotificationService.java
@@ -345,6 +345,7 @@ public class ImageDownloadNotificationService extends Service {
                                             NotificationManager.class);
                             notif.flags |= Notification.FLAG_AUTO_CANCEL;
                             if (mNotificationManager != null) {
+                                mNotificationManager.cancel(id);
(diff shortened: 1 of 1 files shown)


test2a:
I had a feature request before, can we just stop the download notifications altogether ? I use slide a lot but I have to constantly keep swiping those notificcations away

budde25:
@test2a 
That I good point, I for one wouldn't mind getting ride of the downloading notification while still keeping the finished one, or even altogether but I think it would definitely be best to make them options in the settings. Not sure how other people feel. Regardless I think it would still be beneficial to merge this in the meantime.

ccrama:
Thank you for the fix @budde25 !

@budde25 @test2a I'm planning on dropping the notification for Slide v2, in favor of a toast-like popup. The original goal was to let you click the notification to go straight to the image, so I might still add an option in to enable the notification!

Issue #3035, opened 2019-07-12
[Feature Request] When saving pictures, automatically save the post title to the picture file name just like the RES extension.


xCapRx:
Title pretty much says it all. I appreciate how this app automatically saves and organizes pictures into folders with the same names as the subreddits they're sourced from. However, it doesn't seem complete if the pictures aren't named after their parent submission titles. If a picture just has a random string of characters, a lot of **context** is lost. Also, it's a little more of a challenge to manually rename pictures on mobile than it is on desktop PC. 

Since this feature is already a reality in the [open source RES extension](https://github.com/honestbleeps/Reddit-Enhancement-Suite), could it be easily copied or adapted to the Slide app? 

Thanks for your consideration.

Tensounder54:
I've been pushing for a certain amount of unification for a while now. Tis is something to add to the essay I'm writing on feature differences between Slide and RES. I'm hoping to have it done by Jan 2020 with a view to implement the majority of the features in 2020/Slide V2

ccrama:
Fixed by #3297

PR #3297, opened 2020-11-20
Use the submission title as the title for the saved image.

ZianeA:
This feature was requested here: [#3035](https://github.com/ccrama/Slide/issues/3035). This also fixes a bug that causes the application to crash when trying to download an image from a gallery.


ccrama:
Really like the changes, thank you for the PR! Also, thank you for fixing that Gallery bug.

I made a few modifications, and have one future conern:
* Use regex to remove any non-alphanumeric characters from the File URL
* Passing comment's submission title might conflict when downloading future images, if both are from the same thread (might be a really small use case)

Thank you again!

ccrama:
Fixed those concerns with af56312db89077a21738a55148d2b58eee612bec, this will also make sure downloads from subreddits that have repeated titles (r/me_irl comes to mind), it won't overwrite these images

PR #3300, opened 2020-11-28
Fix crash when toolbox config doesn't have values

timawesomeness:
fixes https://www.reddit.com/r/slideforreddit/comments/k29muy/bug_modtools_remove_with_reason_crash/


ccrama:
Thank you for the PR!

PR #3318, opened 2020-12-26
Add High Quality GIFs option (60 FPS)

FLeXyo:
Let me lead off with the fact that I have never touched Android development before, so bear with me.

I noticed that like some other Reddit apps, there is no option for the internal viewer to load 60 FPS gifs from for example Gfycat. I think that this option would be very nice to have (especially since it's so easy to implement), so I took the matter into my own hands and tried to implement it.

Please let me know what you think.


ccrama:
Hey thank you again for the PR! I've released this as part of v6.6.1-4 which should hit the store today :)

PR #3322, opened 2020-12-30
Fix jumping to OP messages in comments

vargon:
Fixes the "sort by OP" functionality which was broken
due to a bad switch/case.


ccrama:
Good catch, thank you for the PR!

PR #3336, opened 2021-02-11
Use submission title as the file name for GIFs / Videos with improved saving file name logic

ni554n:
I made some improvements over #3297.

- Submission title is now safe to use as a file name. It can have all the symbols and emojis. It will be properly sanitized to be saved on any platform.

- File names will be limited to 255 bytes as per Linux requirement. Also handles cases for broken UTF-8 characters due to truncation.

- GIFs and videos will now use submission title as file name, like images.


ccrama:
Great stuff, thank you very much for the PR! Including this in the next update!

PR #3354, opened 2021-03-25
Allow to search by time

carlphilipp:
Hello,
Here is some code that extracts the time parameter from the URI and pass it in the intent for the `Search` Activity.
Let me know if I need to change anything.


ccrama:
Looks good, thank you for the PR!

PR #3360, opened 2021-04-06
Stop changing global dark mode setting on start

### LINKED ISSUE #3065: Night mode turning off on opening a post
Slide version:Slide v6.2.2
Android version: 9.0.0

BrandomRobor:
Fixes #3065

This method is supposed to check if the "Automatic" option of dark mode is supported, but due to the setNightMode() method the setting changes globally. By removing this call the bug no longer happens and the auto check still succeeds.

Photo of "Night mode" settings dialog showing the "Automatic" option correctly:
Screenshot_20210406-152121


ccrama:
Thank you for the PR! Looks good, including this in the next update

PR #3362, opened 2021-04-20
Prevent Slide blanking existing toolbox configs

timawesomeness:
Slide can accidentally blank Toolbox configs if it receives a non-404 error when attempting to load a Toolbox config. 

Fixes https://discord.com/channels/407573578985242635/407573579543347202/834160861114794076


ccrama:
Thank you for the fix @timawesomeness !

PR #3365, opened 2021-04-23
Change some xml-tags attributes

### LINKED ISSUE #3101: Layout params not used
The layout_param is not defined. It should be removed to avoid useless attribute processing at runtime. Found in mod_action.xml

AmooHashem:
In mod_action.xml file, some attributes had problem (such as android:tint="?attr/tintColor" that should be replaced with cardview:tint="?attr/tintColor") that i fixed.

But i didn't find any redundant attributes (according to #3101)! I think you should close this issue.
Fixes #3101


ccrama:
Looking good, thank you for the PR @AmooHashem!

PR #3366, opened 2021-04-23
Replace Hashmap with SparseArray in PeekView.java and HistoryPosts.java

### LINKED ISSUE #3100: Hashmap can be replaced with SparseArray
There are several java classes where Hashmap could be replaced with SparseArray for the sake of performance. For instance:
- PeekView
- HistoryPosts

AmooHashem:
According to issue #3100, for better performance, it's a good choice to replace Hashmaps with SparceArrays.
Fixes #3100


ccrama:
Thank you for the PR! Looks good.

TacoTheDank:
This will be reverted, as it seems to have broken the history and read later features.

PR #3372, opened 2021-04-30
Fix crash with %uXXXX encoded chars in toolbox removal reason fields

timawesomeness:
Fixed this earlier for main removal reason text in #3197, but didn't fix it for header or footer fields, so this fixes that.


ccrama:
Thank you for the PR!

PR #3439, opened 2022-01-03
Update AndroidManifest.xml (should fix #3429)

### LINKED ISSUE #3429: Unable to open "reddit.com" links by default
Slide version: 6.7.1
Android version: 12

After updating to Android 12, Slide is only able to add the following as default supported links:
"amp.reddit.com"
"redd.it"

Attempting to add use the "Add link" option, does not list "reddit.com" as a possible option as shown in the attached image.

image

This makes it so any normal reddit link, found in a google search for example, will always open in the default browser rather than Slide. This was not an issue in Android 11.

cYDN48:
Two changes:
- Delete lines 466 to 487 as they seem redundant
- Change "\*reddit.com" to "\*.reddit.com"

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -463,30 +463,8 @@
                 <category android:name="android.intent.category.BROWSABLE"/>
 
                 <data android:scheme="http"/>
-                <data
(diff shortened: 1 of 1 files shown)


TacoTheDank:
Thank you :)

PR #3473, opened 2022-11-17
Fix redgifs playback

gringofe:
Hi,

this PR fixes redgifs playback by adding the required headers.


TacoTheDank:
Thanks.

bqv:
hey, that's my code! (lmao)

1fexd:
What's your code?

bqv:
i was sure i came up with this fix ages ago, but looks like i might have just been inspired by you tbh so nevermind :D

1fexd:
Actually I'm not sure how I came up with the fix but I think I just looked at the headers whcih are set in the browser

